PR #37, opened 2023-04-05
Warning CSS improvements and illustration of patterns

agjohnson:
I'll see if I can explain some of the smaller changes here next up, but opening this PR for now

Improvements here:

- Reduced the required CSS/HTML structure in 9867509d9e3884afe221bf2e3e8c245f2ba3a262.
  - Dropped extra unneeded divs
  - Used direct descendant specifiers instead of verbose names
  - Dropped unneeded CSS rules for spacing issues
- In the rest of the commits I explored a Lit web component implementation more. Improvements from this approach:
  - Linking HTML, CSS, and JS is much cleaner, and still uses HTML string templates even. String templates could be a point of extension later.
  - Shadow DOM + CSS is much nicer than trying to inject with arbitrary HTML/CSS.
- Started to illustrate some patterns we could aim for
  - CSS variable theming
  - Addon class abstraction for to standardize enabling/disabling (and later configuration) of each addon module
  - Warning variants

I did play around with styles here as well:

image

Here's what the HTML looks like:

```html
<readthedocs-warning><!-- The default class is floated --></readthedocs-warning>
<readthedocs-warning class="inverted"></readthedocs-warning>
<readthedocs-warning class="inverted raised"></readthedocs-warning>
<readthedocs-warning class="inverted" style="--readthedocs-warning-background-color: purple;"></readthedocs-warning>
<readthedocs-warning class="titled raised"></readthedocs-warning>
<readthedocs-warning class="titled inverted raised"></readthedocs-warning>
```


.prettierignore
```diff
@@ -0,0 +1 @@
+*.svg
```
humitos, line 1: 👍🏼

.gitignore
```diff
@@ -1,4 +1,2 @@
 node_modules/
 .direnv/
-
-dist/
```
humitos, line 4: I was ignoring this for now because updating the .js on each commit was dumb at this point of development.

src/non-latest-version-warning.js
```diff
@@ -40,6 +40,9 @@ export function injectNonLatestVersionWarning(config) {
     );
     let node = document.createElement("div");
     node.innerHTML = admonition;
-    main.insertBefore(node, main.firstChild);
+
+    if (main) {
+      main.insertBefore(node, main.firstChild);
+    }
```
humitos, line 46: This change seems unrelated to this PR.

I've changed this pattern also to use `insertAdjacentHTML` because I don't like "creating a wrapper div" as this code does.

src/external-version-warning.js
```diff
@@ -1,63 +1,159 @@
...
+  /** @static @property {string} - registered HTML element tag name */
+  static elementName = "readthedocs-warning";
+
+  /** @static @property {Object} - Lit reactive properties */
+  static properties = {
+    config: {
+      state: true,
+      // Update derived fields from config data
+      // TODO the URLs here should come from a backend API instead
+      hasChanged: (before, after) => {
```
humitos, line 20: I'm not sure to understand where this "after" comes from yet. I'll keep reading, tho.

src/external-version-warning.js
```diff
@@ -1,63 +1,159 @@
...
+
+  renderExternalVersionWarning() {
+    library.add(faCircleXmark);
+    library.add(faCodePullRequest);
+    const xmark = icon(faCircleXmark, {
+      title: "Close",
+    });
+    const iconPullRequest = icon(faCodePullRequest, {
+      title: "This version is a pull request version",
+      classes: ["header", "icon"],
```
humitos, line 74: He, I started removing this "generic" classes because they are pretty common names and they interfered with styles coming from somewhere else.

I understand this doesn't happen anymore with shadow root! 💪🏼

src/external-version-warning.js
```diff
@@ -1,63 +1,159 @@
...
+    // Load this first as it is illegal to instantiate the element class without
+    // defining the custom element first.
+    customElements.define("readthedocs-warning", WarningElement);
 
-    const xmark = icon(faCircleXmark, {
-      title: "Close",
-      classes: ["readthedocs-external-version-warning-close-icon"],
-    });
+    // If there are no elements found, inject one
+    let elems = document.querySelectorAll("readthedocs-warning");
```
humitos, line 135: You are using `readthedocs-warning` here but then in the HTML it's called `readthedocs-banner`. I'm not sure if that is on purpose, or a typo.

I'd prefer to go with the generic one "banner" or even "notification" because we will want to show different messages that are not only warnings.

src/external-version-warning.js
```diff
@@ -1,63 +1,159 @@
...
+ * We load the param ``config`` into the elements after creation and API
+ * response, as this is needed to give hardcoded elements access to ``config``
+ *
+ * @param {Object} config - Addon configuration object
+ */
+export class WarningAddon {
+  constructor(config) {
+    // Load this first as it is illegal to instantiate the element class without
+    // defining the custom element first.
+    customElements.define("readthedocs-warning", WarningElement);
```
humitos, line 132: Where `customElements` comes from? Is it global?

src/warning.css
```diff
@@ -0,0 +1,120 @@
+:host > div {
+  margin: 1rem 0rem;
```
humitos, line 2: With an isolated root, `rem` unit has a lot more sense to me now 💯

src/warning.css
```diff
@@ -0,0 +1,120 @@
...
+  padding-top: 1rem;
+  padding-bottom: 1rem;
+  overflow: auto;
+  border-radius: 0.5rem;
+  font-family: var(
+    --readthedocs-warning-font-family,
```
humitos, line 8: How do people define this CSS variables from outside?

src/warning.css
```diff
@@ -0,0 +1,120 @@
...
+    rgb(234, 234, 234)
+  );
+}
+
+:host(.raised) > div {
+  box-shadow: 0 2px 4px 0 rgba(34, 36, 38, 0.12),
+    0 2px 10px 0 rgba(34, 36, 38, 0.15);
+}
+
+:host(.titled) > div {
```
humitos, line 24: What does `:host` mean here? If I have to guess I'd say that it's the custom element we just defined, `readthedocs-warning`. Is that correct?

humitos (approved):
This is great! I've already mentioned in Slack too.

The pattern is pretty clear and I understand it 90% after watching a 30m video about WebComponents 😄 . That's a good sign in my opinion.

It also clearly solves lot of issues I had to deal with because of colliding with external CSS classes (the search modal has _a lot_ of these). The isolation that this gives us is awesome!

I'm happy to move forward with this pattern and keep exploring in following PRs. I'd like to take a look at `slots` as well so we can allow customization of the HTML in the future. It's not required for now, tho, but thinking about this from the beginning may help us to design a customizable/extendable HTML structure. However, I'm not worried about "breaking changes" at this point in time. Nobody is gonna customize this from the beginning, and we can clearly communicate that's in alpha and things will change.

On the other hand, I don't want to keep talking about "the perfect pattern" forever and freeze the momentum we are having here.

src/external-version-warning.js
```diff
@@ -1,63 +1,159 @@
...
-import { faCircleXmark } from "@fortawesome/free-solid-svg-icons";
+import {
+  faCircleXmark,
+  faCodePullRequest,
+} from "@fortawesome/free-solid-svg-icons";
+import { html, render, LitElement } from "lit";
+
+import styleSheet from "./warning.css";
+
+export class WarningElement extends LitElement {
```
humitos, line 10: One thing that I missed to ask here.

What exactly provides `LitElement` over `HTMLElement`? -- which is what I saw in the video. I understand that `HTMLElement` is the default/standard library when using no framework at all. However, it's not clear yet to me what are the benefits of Lit.

src/warning.css
```diff
@@ -0,0 +1,120 @@
...
+  height: 1rem;
+  vertical-align: middle;
+  cursor: pointer;
+  color: var(--readthedocs-warning-title-color, rgba(96, 96, 96));
+  font-weight: normal;
+}
+
+:host(.inverted) > div > .title > .right > svg {
+  color: var(--readthedocs-warning-title-color, rgba(255, 255, 255, 0.9));
+}
```
benjaoming, line 120: Maybe you'll end up pretty soon adding media queries for responsive layout stuff and dark themes.

I can only imagine that path leading towards SCSS as well :)

```
@media (prefers-color-scheme: dark) {
  @extend .inverted;
}
```

src/external-version-warning.js
```diff
@@ -1,63 +1,159 @@
...
+  static elementName = "readthedocs-warning";
+
+  /** @static @property {Object} - Lit reactive properties */
+  static properties = {
+    config: {
+      state: true,
+      // Update derived fields from config data
+      // TODO the URLs here should come from a backend API instead
+      hasChanged: (before, after) => {
+        if (after && Object.keys(after).length) {
```
humitos, line 21: What `Object.keys` does? Does it return all the keys from that object?

src/external-version-warning.js
```diff
@@ -1,63 +1,159 @@
...
+    library.add(faCodePullRequest);
+    const xmark = icon(faCircleXmark, {
+      title: "Close",
+    });
+    const iconPullRequest = icon(faCodePullRequest, {
+      title: "This version is a pull request version",
+      classes: ["header", "icon"],
+    });
+
+    return html` <div @click=${this.closeWarning}>
```
humitos, line 77: This `@click` should be tied to the X icon. Otherwise, the banner closes when you click on any part of it, even in the text if you want to select it.

Another option is to check what's the element click in the `closeWarning` handler.

agjohnson: We can move this now that the remove method is calling `this.remove()`, though because we can't alter the icon attributes, putting this on the right float wrapper is probably the closest.

However, I also don't mind the interaction of just clicking anywhere on the element

src/external-version-warning.js
agjohnson: Correct, this is checking to see if the object is empty, as `if (after)` does not correctly catch an empty object. This is one of a number of places where the stdlib for JS is woefully minimal. Even this is not technically correct, there are _additional checks required_ to avoid different object types.

src/warning.css
agjohnson: Yup! Next step somewhere is adding LESS. I didn't want to go down that hole just yet though, but it would clean up the CSS a good deal.

src/warning.css
agjohnson: Yup! The style docs describe this in a little detail, but I think it's also CSS spec? https://lit.dev/docs/components/styles/#host

src/warning.css
agjohnson: Authors can define these variables on `:root` or `readthedocs-warning` selector, so:

```css
readthedocs-warning, :root {
  --readthedocs-warning-background-color: purple;
}
```

src/external-version-warning.js
agjohnson: Ahh yes, it's global and from the browser, similar to window/document.

src/external-version-warning.js
agjohnson: Yeah, I was going to bring up a conversation here. I like "notification" or "message" over "warning". I started calling in banner, but then made it possible to inline it, so banner made less sense :slightly_smiling_face: 

I think a few points to discuss are:

- Is this a combination of external warning and version outdated warning? I feel like we'll want just one block for both of these messages, as they also can't both show at the same time.
- But then, do we also want to show other messages in this block too? I think this is where I'd absolutely agree with a more generic name
- Do we think there will be more than one message showing at a time? This would change the floating implementation a little bit

src/external-version-warning.js
agjohnson: It does a great job cleaning up the CSS. I stuck with a pattern like SUI because we also don't need to worry as much about overlapping class names now.

src/external-version-warning.js
agjohnson: This comes from https://lit.dev/docs/components/properties/#haschanged

When we have URLs in the config data, this particular block won't be necessary anymore. However, it's likely we might want to update dependent properties using this pattern elsewhere though

src/non-latest-version-warning.js
agjohnson: This was another exception being throw, and I believe it was causing some problems executing the warning code perhaps? This is another issue that would be solved with checks before the addon is executed.

.gitignore
agjohnson: Makes sense! We can comment out some things for now in CI/etc

src/external-version-warning.js
humitos: > but then made it possible to inline it, so banner made less sense

What does it mean in this context?

src/warning.css
humitos: Yeah, please. One step at a time 😄

src/external-version-warning.js
agjohnson: > What does it mean in this context?

"Banner" to me implies a floating element, where users can inline the element now (that is, not a fixed position but a inline/inline-block position).

> We should not show both at the same time. It doesn't make sense.

Cool, that's what I assumed. It seems like we could put both in the same warning/banner/message/etc then, unless we're thinking we want them separate for another reason.

> I'd like to have a "generic banner" where we can change the content and the icon of it at instantiation time, for example.

Same. I don't yet know what that is, but I suspect we might eventually come up with something else to add.

> I don't think so, no. At least for now. We can adapt the implementation later if this grows in that direction, or create a new "notification type" for the new ones.

:+1: I assumed this would be the case, especially if the two warnings never overlap

src/external-version-warning.js
agjohnson: That's all true. I will say that "hit the tiny X" is also a frustrating UX though, especially on mobile. There is probably not a great middle ground here unless we detect mouse dragging or selection.

humitos: Mobile can perform differently for sure. It could close when tapping the warning itself, or the X detecting rectangle could be bigger as well. For desktop I think just the X is 💯

src/warning.css
agjohnson: Added at https://github.com/readthedocs/readthedocs-client/issues/39 along with responsive styles as some good next steps

src/external-version-warning.js
agjohnson: Aye we can play with this more later, and maybe play around with mobile events or media queries altering JS behavior here or something. I have a commit that moves the click event to the right float for now though

humitos:
I'm merging this PR for now and probably mine as well. We can create more PR to tweak the details we have been discussing. Nothing is written in stone yet and we have been pretty flexible with the decisions. So, 👍🏼 on keep moving forward.

agjohnson:
I have updates here actually. Hold off for a minute and I'll merge

src/external-version-warning.js
agjohnson: For now, I've dubbed this `notification` instead, in [9f63424](https://github.com/readthedocs/readthedocs-client/pull/37/commits/9f63424e6a5e9863928e163c8551e6a3ce96137f)

src/search.js
```diff
@@ -681,6 +681,10 @@ function getCurrentFilter(config) {
 }
 
 export function initializeSearchAsYouType(config) {
+  if (!config.features || !config.features.search) {
+    return false;
+  }
```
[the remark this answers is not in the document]
agjohnson: The pattern I was hinting at is here:

https://github.com/readthedocs/readthedocs-client/pull/37/commits/9f63424e6a5e9863928e163c8551e6a3ce96137f#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R49-R63

This sort of pattern helps ensure the addon is not executed if it does not pass a check for being enabled first. I know most the addons have checks now, an explicit pattern like this is more consistent.

agjohnson:
Okay, addressed a few things we talked about here and a few points of clean up I wanted to merge up before they got lost. Thanks for waiting!

I'll merge everything up and we can iterate from there :+1: If you have any questions to go over though, feel free to ask!

src/utils.js
```diff
@@ -22,5 +27,16 @@ export const domReady = new Promise((resolve) => {
...
 });
+
+/**
+ * Addon base class
+ *
+ * Provides a common interface for addon configuration testing, customization,
+ * and loading.
+ */
+export class AddonBase {
+  static isEnabled() {
```
humitos, line 39: Is it OK to have a `isEnabled()` and `isEnabled(config)` in the same class?

src/index.js
```diff
@@ -31,22 +31,35 @@ export function setup() {
...
             })
           );
         }
 
+        // TODO migrate the above to use a common pattern for addon injection.
+        // Addons should not execute or inject anything if they are not enabled.
+        const addons = [notification.NotificationAddon];
+
+        for (const addon of addons) {
+          if (addon.isEnabled) {
```
humitos, line 54: Won't this always return `false` since it does not receive the `config` object which is required to know if the addon is enabled or not?

humitos (commented):
LGTM

src/utils.js
agjohnson: Good catch. Technically, it's fine, JS is very loose on function argument definition, unlikely Python. But for code cleanliness it should match.

PR #38, opened 2023-04-05
Search: use Lit WebComponents

humitos:
I'm following the pattern from #37 by myself to test out the framework and start to feel more comfortable.


humitos:
I re-implemented the search addon using Lit. I created an element named `<readthedocs-search>`. It's not too configurable right now, but we will get there at some point, creating styles and being able to pass different values, like custom filters and more.

The good news is that it works perfectly! 💯

I was able to simplify the logic we used to have _a lot_, making usage of Lit features but also the CSS because of the shadow DOM 🎉 . I'm pretty happy with the results and confident that it works flawlessly 😉 . I left some `// TODO` comments that we can do in the following iterations.

@agjohnson I don't think it makes sense to _review_ this PR since it changes the old code 100%. I think it's easier to just take a look at the raw file if you are interested into, tho. In any case, I think I will just merge this PR and attack any change in future and smaller PRs.

Here you have an example of how it's working right now.

Peek 2023-04-13 17-10

humitos:
This work is done for now. I'll merge it so we can move forward with this and open smaller PR with the required changes. It does not make too much sense to review the changes, since the implementation changed completely 🤷🏼 

I'm happy to receive feedback via comments on the PR, issues and even code changes via PRs if you want.

PR #45, opened 2023-04-12
Notification: copy and URLs changes

humitos:
- ~~Invert order of arguments for `hasChanged`: https://lit.dev/docs/components/properties/#haschanged~~ `hasChanged` function was removed
- Change the copy from the content to be different from the title and easier to understand what the links mean
- Reduce the `font-size` for the content

Example,

Screenshot_2023-04-12_13-10-18


src/notification.js
```diff
@@ -18,13 +18,17 @@ export class NotificationElement extends LitElement {
...
-      hasChanged: (before, after) => {
-        if (after && Object.keys(after).length) {
+      hasChanged: (after, before) => {
+        if (after && after !== before && Object.keys(after).length) {
           this.urls = {
             build: `${window.location.protocol}//${after.domains.dashboard}/projects/${after.project.slug}/builds/${after.build.id}/`,
             external: `${after.project.repository_url}/pull/${after.version.slug}`,
           };
+            // FIXME: for some reason the element is not being updated once the URLs are defined here.
+            // Resulting in empty URLs on the links :/
```
humitos, line 28: I suppose the `this` inside this function is not what we are expecting to be, probably.

humitos:
I ended up removing the `hasChanged` complexity for now because it's not required for our current use case. The addon is initialized with the whole `config` object and it does not change after that; requiring no refresh render again. So, I'm changing the `this.urls` field insie the `loadConfig` for now which makes the WebComponent to render properly with the correct URLs.

src/notification.css
```diff
@@ -85,6 +85,7 @@
 }
 :host > div > div.content {
   line-height: 1rem;
+  font-size: 85%;
```
agjohnson, line 88: This is where I'd use `em` units, I rare use percent units in fact. I didn't mix `em` and `rem` in my change, but this is totally fine too. The `em` unit works great for this case, where the description line should be proportional to the main block font size.

src/notification.js
```diff
@@ -46,6 +34,17 @@ export class NotificationElement extends LitElement {
...
+    // TODO: this URL should come from the backend API.
+    // Doing a simple replacement for now to solve the most common cases.
+    const vcs_external_url = config.project.repository_url
+      .replace(".git", "")
+      .replace("git@github.com:", "https://github.com/");
+
+    this.urls = {
+      build: `${window.location.protocol}//${config.domains.dashboard}/projects/${config.project.slug}/builds/${config.build.id}/`,
+      external: `${vcs_external_url}/pull/${config.version.slug}`,
+    };
```
agjohnson, line 47: This is comparable to hasChanged and close to what we want effectively, both are fine options :+1:

src/notification.js
```diff
@@ -86,11 +85,13 @@ export class NotificationElement extends LitElement {
...
-          from a pull request (<a href="${this.urls.external}"
-            >#${this.config.version.slug}</a
-          >).
+          Take a look at the
+          <a href="${this.urls.build}">build details' page</a>
+          and/or the
+          <a href="${this.urls.external}"
+            >pull request #${this.config.version.slug}</a
+          >
+          for more context.
```
agjohnson, line 94: Just some copy tuning, this is a great improvement though :+1: 

```suggestion
          See the
          <a href="${this.urls.build}">build's detail page</a>
          or
          <a href="${this.urls.external}"
            >pull request #${this.config.version.slug}</a
          >
          for more information.
```

agjohnson (approved):
Seems we were on the same page, I almost tuned this more too. This is a great improvement.

src/notification.js
humitos: `hasChanged` is good -- However, I removed because it was not working as we expected:

- the arguments order was wrong
- after fixing that, the URLs were still `null` 🤷🏼 

I debugged it a little and didn't arrive at good place. So, I removed the Lit magic and done it in the simplest way 😅

src/notification.css
humitos: Makes sense. Changed to `0.85em`

PR #48, opened 2023-04-12
Fix init pattern bug and clean up base code

agjohnson:
Some fixes @humitos noted post-review


src/analytics.js
```diff
@@ -86,6 +86,6 @@ export class AnalyticsAddon extends AddonBase {
   }
 
   static isEnabled(config) {
-    return this.config.features && this.config.features.analytics.enabled;
+    return config.features && config.features.analytics.enabled;
```
agjohnson, line 89: You were correct on your note about passing `config` into this static method.

Because this is a static method, `this` doesn't refer to an instance of this class, but to the class itself. So, `this.config` won't exist when we call `AnalyticsAddon.isEnabled(config)`.

JS static methods are similar to Python classmethod functions

humitos (approved):
Great!

PR #49, opened 2023-04-12
Fix font family

agjohnson:
Firefox seemingly handles font-family different than Chromium. Firefox
 is okay with a single string, Chromium treats a single string as one
 big font family name and shows a default serif font instead.


humitos (approved):
Ohhh, I see. This makes sense now and it works as expected 💯

Issue #26, opened 2023-03-30
Scope "shippeable" v1 of this js client


humitos:
What's is required to ship a v1[^1] of this client and expose to users using `build.commands`, which currently have 0 integrations with Read the Docs?

After some talks, we started to treat each feature as a _separated feature_ that can be deployed without deploying the "whole flyout". In fact, we decided we want to deploy other features first _without_ a flyout at all.

## Functionality

There are some features that require zero configuration from the user at this point (they can always be expanded to be customized, of course, but we are not there yet. See #13 for that)

Some of these features are:

| Feature | Description | Data required | Triggered by |
| ------- | ----------- | ------------- | ------------ |
| Analytics | Hits an API endpoint to show analytics in the dashboard | Project (slug, language, programming language), Version (slug), Global analytics code | Automatic (always) |
| External version warning | Shows a warning banner if the version is external | Version (type), Domain (dashboard), Project (slug, repository_url), Build (id) | Automatic (when the version is external) |
| Search as you type | Presents a modal to search using our own backend | Project (slug) | Manual (by hitting `/`) |

### Where the data required will come from?

* I'm currently using the new endpoint `/_/readthedocs-config/?url=` with a data structured that was not thought completely and returns a bunch of duplicated data for now
* We could polish it a little to return the real objects and define it in an extendable way so we can include more data in the future
* The [actual structure proposes](https://github.com/readthedocs/readthedocs.org/blob/f141a8f931ffe78d6aad90a4710b7ec566994dcd/readthedocs/proxito/views/hosting.py#L35-L136) an idea for this. In particular using:
  * Top level keys for "generic" data, like `project`, `version`, `domain` objects that are not tied to a specific feature
  * Feature-specific data under `features.<feature>`, like `features.external_version_warning`, `features.analytics`, etc

### CF worker that doesn't randomly 500

We are not sure why, but our script running in CF sometimes it break for some reason and returns an error 1/2 the times we hit it. We will need to debug this.

### Return `X-RTD-Hosting-Integrations` on versions with `build.commands`

We will need to modify the Proxito middleware to check if the version is using `build.commands` and inject the header in that case so CF includes our client.

## Frontend

It would be good if I can get some review on the current HTML structure of these features. I can make the required adjustments to it. However, I would avoid getting too deep into creating _the perfect structure_ since we are not going to officially allow people to style/customize/change them for now.

[...]

PR #52, opened 2023-04-17
Shippeable v1

humitos:
- Analytics
- PR build notification

The API returns `addons.search.enabled` and `addons.non_latest_version_warning` as `true` currently. So, we can decide to re-deploy the JS file enabling this features without doing an application deploy again.

However, for now, we only want these pretty small addons to start testing the architecture and others.

Related #26


agjohnson (approved):
This seems fine for now. If we need to disable more features we're actively developing, it could help to instead gate these features behind conditional checks for Webpack local development or something similar.

humitos:
How that would be? Do you have a documentation link I can read? 😅

agjohnson:
Sure! I would use DefinePlugin: https://webpack.js.org/plugins/define-plugin/

There is a feature flag example there even, though we probably want to vary that value on whether the build was a development or production build, ie https://github.com/readthedocs/readthedocs-client/blob/main/webpack.config.js#L26

Anyways, not something that needs to hold this PR up, but would be an easy later change.

PR #55, opened 2023-04-19
Packaging: define `IS_PRODUCTION` global variable

### LINKED ISSUE #54: Development: pattern to skip addons on production
We may want to have some addons installed while doing development, but not on production.

Take a look at https://github.com/readthedocs/readthedocs-client/pull/52#issuecomment-1512401438

humitos:
Use webpack's `DefinePlugin` to define `IS_PRODUCTION` global variable. This variable is used to skip some features we are still working on and are not ready for production yet.

We want them enabled in development (`npm run dev`) but skipped in production (`npm run build`).

Closes #54


agjohnson (approved):
Looks good!

Issue #13, opened 2023-03-20
User API: communicate with the library


humitos:
There are many integrations that will require some user configuration. For example:

* HTML templates: there are some integrations that use HTML templates to render (e.g. flyout, non latest version warning, external version warning). We need to design a way users can tell the library what HTML template to use or giving them a way to replace ours completely (as flyout is doing via `#readthedocs-embed-flyout`) [^1]
* Functions with logic: [we are using `semver` to decide if the current version is the highest](https://github.com/humitos/readthedocs-client/blob/d3472c417b333b1698b77135f3f8cce4e881ba81/src/non-latest-version-warning.js#L10-L13). However, the user may be using CalVer or a custom versioning pattern. We should allow users to tell us what's the _"function that returns the highest version from a list of versions"_ that we should execute.
* Element query selectors:
  - what's the selector for the _main content_?
  - what's the content selector to apply tooltips to the inner links?
  - what's the selector for the search-as-you-type be triggered?
  - etc 
* Data for functions:
  - default filter for search
  - extra filters for search
* Tooltips:
  - should we apply them to all the content?
  - should we only apply them on a specific selector? (as we currently do with `.hoverxref` class)
  - should the doctool generate HTML tags with this specific selector?
  - should we try a defined selector and fallback to the main content if not found?

Note there may be more things to consider here, but I wanted to mentioned the ones that I found while working on this library. I'd like to define **the pattern** first, since once we have that defined we can easily add more. I'm looking for _"the standard pattern to communicate with a js library from the browser"_, basically.

Also, it's worth to mention that _some of them_ can be converted to conventions (e.g. some selectors) but we need to discuss and agree on them anyways.

[^1]: this work may require writing a policy that users should follow since they can overwrite our templates completely. We may want to "enforce" projects to show _"Hosted by Read the Docs"_ or similar in the flyout, for example.

humitos:
There were some progress in this direction. I'd say, mainly in my understand of what we need from the user and how this could work in the js world. With the addition of WebComponens, CSS variables and others it's a lot more clear to me how people will interact with this library in some of the required ways. There is no decided made already about this, but I guess that:

- we will expose some CSS variables
- users can define CSS variables if they want to change a particular style from the UI
- users will use `data-` attributes or specific CSS classes to specify pre-defined styles for our addons
- we will define some custom WebComponents
- people can place the HTML component whenever they want in the HTML; if there is one present, we won't include "the default one"
[...]

PR #56, opened 2023-04-19
HTML attributes API: configure the addon via HTML attributes

humitos:
By default the search modal is shown when pressing `/`. However, we want this to be configurable by the user. This can be achieved by passing `show-modal-keycode` attribute to the web component as follows:

```html
<readthedocs-search show-modal-keycode="220">
</readthedocs-search>
```

In this case, it will be shown when pressing `\`.

I think this is a good pattern for static values that won't be changed from the dashboard or similar in a dynamic way.

Related #13


agjohnson (commented):
This is what I was thinking for a first pass too, this patterns seems like a good fit for us :+1:

humitos:
Thanks for the feedback. I updated the code to use Lit properties.

agjohnson (approved):

PR #58, opened 2023-04-20
Search: define which element & event shows the modal

humitos:
Define two new attributes `show-modal-selector` and `show-modal-event` to allow the user defining which HTML element will show the modal.

Peek 2023-04-20 09-45

In this example, when clicking on the `input` it will show the modal.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -20,7 +20,13 @@
 
     <p>Nothing to see here, just a placeholder to put the <code>readthedocs-search</code> tag.</p>
     <p>Currently, it modifies the tag to be shown when pressing <code>\</code> (keycode 220) instead of <code>/</code> (keycode 191)</p>
-    <readthedocs-search show-modal-keycode="220"></readthedocs-search>
+    <readthedocs-search
+        show-modal-keycode="220"
+        show-modal-selector="#search-input"
+        show-modal-event="focusin"
+    ></readthedocs-search>
+
+    <p>Clicking in the following input should show the search modal: <input id="search-input" placeholder="Search docs" /></p>
 
     <h2>Notification</h2>
 
--- a/public/index.html
+++ b/public/index.html
@@ -21,9 +21,8 @@
     <p>Nothing to see here, just a placeholder to put the <code>readthedocs-search</code> tag.</p>
     <p>Currently, it modifies the tag to be shown when pressing <code>\</code> (keycode 220) instead of <code>/</code> (keycode 191)</p>
     <readthedocs-search
-        show-modal-keycode="220"
-        show-modal-selector="#search-input"
-        show-modal-event="focusin"
+        trigger-keycode="220"
+        trigger-selector="#search-input"
     ></readthedocs-search>
 
     <p>Clicking in the following input should show the search modal: <input id="search-input" placeholder="Search docs" /></p>
--- a/src/search.js
+++ b/src/search.js
@@ -41,9 +41,9 @@ export class SearchElement extends LitElement {
       state: true,
     },
     cssFormFocusClasses: { state: true },
-    showModalKeycode: { type: Number, attribute: "show-modal-keycode" },
-    showModalSelector: { type: String, attribute: "show-modal-selector" },
-    showModalEvent: { type: String, attribute: "show-modal-event" },
+    triggerKeycode: { type: Number, attribute: "trigger-keycode" },
+    triggerSelector: { type: String, attribute: "trigger-selector" },
+    triggerEvent: { type: String, attribute: "trigger-event" },
   };
 
   static styles = styleSheet;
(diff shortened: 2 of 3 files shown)


humitos:
This is another example towards the definition of the pattern we want to implement. It answers the question **what's the selector for the search-as-you-type be triggered?** from #13 

I'm not 100 sure about this pattern yet, but I opened this PR as an example and to start receiving feedback. However, I think it's simple enough and easy to customize and maintain, I think.

src/search.js
```diff
@@ -490,13 +494,28 @@ export class SearchElement extends LitElement {
     }
   };
 
+  _handleShowModalUser = (e) => {
```
agjohnson, line 497: Any reason this isn't a normal class method instead of an anonymous function? It seems to be used as a method, given `this.showModal()`

public/index.html
```diff
@@ -20,7 +20,13 @@ <h2>Search</h2>
 
     <p>Nothing to see here, just a placeholder to put the <code>readthedocs-search</code> tag.</p>
     <p>Currently, it modifies the tag to be shown when pressing <code>\</code> (keycode 220) instead of <code>/</code> (keycode 191)</p>
-    <readthedocs-search show-modal-keycode="220"></readthedocs-search>
+    <readthedocs-search
+        show-modal-keycode="220"
+        show-modal-selector="#search-input"
+        show-modal-event="focusin"
```
agjohnson, line 26: I was going to comment on the last PR too, but figured it was a one off attibute. But these attributes probably don't need to mention "modal", as that is an implementation detail of sorts on our side. That is, users are showing the RTD search, not just a modal -- plus modal leans towards jargon.

Also, thinking way ahead, I could see us moving away from a modal, as it is rather disruptive to the flow/look/feel of the site. It's still an ideal place to start though.

So, maybe naming should be closer to `trigger-selector`, `watch-selector`, etc?

src/search.js
```diff
@@ -65,6 +67,8 @@ export class SearchElement extends LitElement {
     this.defaultFilter = null;
     this.filters = [];
     this.showModalKeycode = 191;
+    this.showModalSelector = null;
+    this.showModalEvent = null;
```
agjohnson, line 71: We should probably set a default event of "focusin". This allows for users to just define the selector. Right now, both fields are required and if you only define one nothing will happen.

src/search.js
```diff
@@ -490,13 +494,28 @@ export class SearchElement extends LitElement {
...
+    e.preventDefault();
+    this.showModal();
+  }
+
   connectedCallback() {
     super.connectedCallback();
     // open search modal if "forward slash" button is pressed
     document.addEventListener("keydown", this._handleShowModal);
+
+    if (this.showModalEvent && this.showModalSelector) {
```
agjohnson, line 507: If we define a default for the event type, this logic doesn't need to change. But the default for the event type will only require the user defines the selector attribute, instead of always having event be a dependent field.

agjohnson (commented):
I think this is an good pattern for us to have, though I can also see how more advanced maintainers will want something more direct, like calling a JS method directly. That can be a later addition, this pattern will work for the majority of users.

src/search.js
humitos: Yes. There is a reason for this,

https://github.com/readthedocs/readthedocs-client/blob/3e28931eb0eee6d198a04c9d43416dc5f03dceb6/src/search.js#L482-L483

public/index.html
humitos: `trigger-` works a prefix to me 👍🏼

src/search.js
agjohnson: Gotcha, that makes sense. I was going to suggest a static method, but I think in that scope `this` refers to the class itself.

src/search.js
```diff
@@ -64,7 +66,9 @@ export class SearchElement extends LitElement {
     this.currentQueryRequest = null;
     this.defaultFilter = null;
     this.filters = [];
-    this.showModalKeycode = 191;
+    this.triggerKeycode = 191;
```
humitos, line 69: @agjohnson should we have a `trigger-modifier` here or similar? See https://github.com/readthedocs/readthedocs-client/issues/57

How can we make people to use `Ctrl + K` for example to show the search?

humitos:
I think this is ready to merge. I asked a question, but it can probably be achieved in a different PR. Let me know.

src/search.js
agjohnson: Good question. I haven't done this, but it looks like we can get the modifier state from the keydown event passed into our event handler function.

I will say that we should be careful with defaults here though. It's easy to override native key bindings when working with modifier key presses.

agjohnson (approved):
Yup, looks good!

src/search.js
humitos: Is it fine to add a `trigger-modifier-keycode` attribute for this? 🤔

agjohnson: Yeah, something like that, though I don't think `keycode` is the most correct. I believe this is the KeyEvent method you'd want to use:

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState

So, more like `<readthedocs-search trigger-key-modifier="control">` or something.

Issue #53, opened 2023-04-17
"Tool bar" or "Dashboard" for addons


humitos:
We have been talking about a way, as a reader, to _Enable/Disable_ addons. There are situations where it would be pretty useful to disable `doc-diff`, for example.

I'd like to propose to _decouple_ this interaction from the flyout if possible [^1]. Right now, the flyout serves its own purpose showing published versions, offline format links and others.

I was thinking we probably want something more similar to "Django Debug Toolbar" for this purpose:

| Collapsed | Expanded |
| ---------- | ---------- |
|  ![Screenshot_2023-04-17_15-59-37](https://user-images.githubusercontent.com/244656/232507786-77d02e95-1489-4a90-b490-d6f104fd20b9.png) | ![Screenshot_2023-04-17_15-59-50](https://user-images.githubusercontent.com/244656/232507805-975e0ecd-93a0-4ca1-9a88-08d22d7021ac.png) | 

Note that something simple like this could already serve our own purpose as well, since we could _activate/de-activate_ addons via checkboxes, similar to what "Django Debug Toolbar" does. We could do a PoC creating a `<readthedocs-toolbar>` element with some simple checkboxes and see how it goes.

In any case, I'm opening this issue here to start collecting ideas that people may already have, discuss them, and build a quick PoC that we can start testing out.

[^1]: Maybe, what we need here is to re-think the flyout completely again and re-design it 🤔 . However, that will take a lot more time than a PoC, I'd guess 😄

humitos:
There are other features we have talked about some time about that are not "Enable/Disable" things that we could consider adding here as well. I just found that we talked about "Share this page" link on RTD4B that automatically generates a secret link to that page: https://github.com/readthedocs/readthedocs-corporate/issues/189

humitos:
After talking with Eric at PyCon and thinking more about this, I'm imagining there is going to be **a new flyout** completely redesigned that will host all these addons.

Similar to what we currently have, but _not natively integrated with the theme_, _always visible_ and showing _only_ the Read the Docs logo when collapsed [^1]. I think this are key properties that Read the Docs should provide as a hosting platform so all the readers have exact the same experience when surfing documentation at Read the Docs.

Even if authors integrate our addons in a "better way", we will still always provide our own flyout because the user will find all the addons there and it will always look the same -- no matter what's the theme/doctool the original author is using to build their site. This is a vague example of what I'm thinking about:

|Flyout bottom right|Flyout right|
|----|----|
[...]

PR #63, opened 2023-04-25
Docdiff: initial work to toggle enabled/disabled

### LINKED ISSUE #11: Docdiff: investigate why it breaks other integrations
Docdiff works correctly currently. However, I had to disable it for now because it interferes with other integrations like EthicalAds and Search As You Type.

I assume that it breaks because it changes the DOM and it may be removing some things required for the other integrations? 🤷🏼

### LINKED ISSUE #23: Docdiff: integrate nicely
_Description in progress (I'll be editing it)_

How do we want to integrate with docdiff?

- enable by default on PRs
- ability disable and re-enable by the reader
- as a reader, how I can decide how to diff? show only additions, show only removals, both?
- etc

References: https://github.com/readthedocs/meta/discussions/38

### LINKED ISSUE #62: Make `<readthedocs-docdiff>` an addon
https://www.w3schools.com/howto/howto_css_switch.asp

Screenshot_2023-04-24_16-43-29

I'd like to have this "toggle" accessible for theme authors to be able to put in their on page. We can think about adding it to the flyout. See #53

humitos:
Peek 2023-04-25 16-43

Closes #62 
Closes #23 
Closes #11


humitos:
I added a smaller example in the `public/` folder so we can test it locally and play with it.

Peek 2023-04-27 15-49

humitos:
Note that I'm not putting this anywhere in the page automatically for now. I think this will probably be part of #53 instead.

humitos:
I've used `Control + Shift + f` as the hotkey for now. I'd appreciate other ideas here, but we will need to discuss about a more general "hotkey pattern" we want to follow with all our addons. It works great for now:

Peek 2023-07-13 15-52

humitos:
I'd like to have a review here so we can merge it and start testing it in our projects at least.

humitos:
> I've used `Control + Shift + f` as the hotkey for now

I reverted this decision and use single-stroke `d` after researching a little more about this in #80

agjohnson:
Just a note here for now, but I could see this implementation drifting closer to either a multiposition slider, pair of toggle bars, or even just a dropdown instead (I lean towards this). There are a few modes that I'm finding useful, depending on the pull request contents:

- On, show deletes and additions
- On, only show additions
- Off

This doesn't need to affect the direction here yet, but will be something we want to consider in the future. Just noting so we don't cement too much of our implementation around a single, binary toggle bar. For the purpose of turning the diff off, this likely works fine for now.

humitos:
I'm going to merge this because it's working fine. The only way to discover this feature right now is by hitting `d` when viewing a Pull Request. I will be promoting slowly this to start getting feedback.

We can easily disable from the API response if we notice it breaks for some reason.

PR #64, opened 2023-04-27
API: use `<meta>` to define supported API version and trigger `CustomEvent`

### LINKED ISSUE #17: Forwards compatibility (error handling)
The overall question is how do static websites work in 5-10 years and how many legacy APIs do we expect to be around by then.

The solutions are perhaps found in some sense of user-oriented error handling.

* Does the API respond? What if it doesn't?
* _When_ were the docs generated (at what date)? Do we have a customized message in case docs are _really old_ and the API isn't around?
* Can the version warning banner help?
* Are reasonable fallbacks possible? What reliable resources can help when others are failing? For instance, a link to the "latest version" or the project's page on Read the Docs?

https://github.com/readthedocs/readthedocs.org/issues/8323 is an example of assumptions that may have broken after 5-10 years.

**_This is possibly pretty low priority, just wanted to post these thoughts somewhere :)_**

### LINKED ISSUE #60: Trigger a `CustomEvent` on config load
This would be useful for theme authors

https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent

### LINKED ISSUE #61: Let theme authors decide the version of the data supported
Add a `<meta name="readthedocs" content="version: 1.0" />`

This will let the client how to expose the `window.readthedocs` object. The client may need to do an extra call to the API to get the data in the structure the theme is expecting.

See #60 which is the one that will use this to keep compatibility.

humitos:
Let users to define `<meta name="readthedocs-api-version" content="1.0">` to tell Read the Docs client what is the API scheme version supported by them.

When our client request the API data, if the `api_version` returned does not match with the one expected by the user, another request is done to force a particular API scheme version.

Then, we dispatch a `readthedocsdataready` custom event ~~and expose `window.readthedocs`~~, to let our users know this data is ready to be consumed by their own integrations.

Closes #60
Closes #61
Closes #17 
Closes https://github.com/readthedocs/readthedocs.org/issues/9957
Closes https://github.com/readthedocs/addons/issues/250


src/readthedocs-config.js
```diff
@@ -19,11 +31,56 @@ export function getReadTheDocsConfig() {
...
+        if (metadataAPIVersion !== data.api_version) {
+          // When the API scheme version returned doesn't match the one defined via `<meta>` tag by the user,
+          // we perform another request to get the Read the Docs response in the structure
+          // that's supported by the user and dispatch a custom event letting them know
+          // this data is ready to be consumed under `event.detail` or `window.readthedocs`.
+          fetch(url, {
+            method: "GET",
+            headers: {
+              "X-RTD-Hosting-Integrations-Version": CLIENT_VERSION,
+              "X-RTD-Hosting-Integrations-API-Version": metadataAPIVersion,
```
humitos, line 56: This is a new header that has to be read from the endpoint, in a similar way as we are doing at https://github.com/readthedocs/readthedocs.org/blob/40d21dc6cc9378b4cd6d274f02d043c01da510da/readthedocs/proxito/views/hosting.py#L58

This header will force the response to be in a particular scheme (the one supported by the user/theme/etc), instead of the "latest supported by the js client".

src/readthedocs-config.js
```diff
@@ -19,11 +31,56 @@ export function getReadTheDocsConfig() {
...
+                return undefined;
+              }
+              return response.json();
+            })
+            .then((data) => {
+              const readthedocsDataReady = new CustomEvent(
+                "readthedocsdataready",
+                { detail: data }
+              );
+              window.readthedocs = data;
```
agjohnson, line 71: What is the purpose of setting `window.readthedocs`?

Usually, data related to an event is only passed in on the Event instance, like you're doing above with the `detail` attribute. I'd probably lean towards just having the singular pattern of using an Event, instead of mixing evented and global state patterns.

humitos: The purpose here was to allow users to access this data as `window.readthedocs.addons...`. I'm not 100% sold on this, but I added because we talked about exposing this object in multiple cases. I felt that _always using even driven data_ may not cover all the cases, but maybe it's enough.

I'm happy to remove this if we can't find a good use case. We can always add it later if we have one.

agjohnson: There is, or will be, a good use case for accessing the library directly like this, but I think we can still make this 100% evented. Right now, I believe the only global state this library uses is when web components are registered. Otherwise, it's fully isolated by Webpack.

I'd say lets aim for 100% evented, and we should attach what we think the user needs to the Event instance. IE:

```js
document.addEventListener("readthedocsdataready", function(event) {
  console.dir(event.readthedocs.addons);
  // Or maybe
  console.dir(event.addons);
  // API response
  console.dir(event.response);

  return False
});
```

I could also see more events too, like maybe there is a `readthedocsready`, etc. We can think about others to add later though.

humitos: > I'd say lets aim for 100% evented, 

Cool. I will remove the `window.readthedocs` for now, then.

> and we should attach what we think the user needs to the Event instance.

This is done already. I'm passing the whole API response in `detail` (see https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#parameters). Then users can already use it like in the example I put inside the `index.html` file.

humitos:
@pradyunsg 👋🏼 -- I'm pinging you in this PR because this is the initial implementation for the pattern that we talked at PyCon to keep the integration stable as much as possible. I'd appreciate any feedback you may have here.

Let me know if connecting to the `readthedocsdataready` event as a theme author and having access to API response as JSON in `event.detail` would be enough for you.

Feel free to suggest any change you consider, from naming to full re-designing of the pattern 😉

src/readthedocs-config.js
agjohnson: Yeah, I noted the `detail` attribute above, but that's only the API response data, no? If we want to give access to something like `readthedocs.addons`, that won't be in the API response data.

I might suggest something other than `detail` as an attribute to make it a bit more clear what the structure is.

humitos: Well, the idea of giving users access to the API response data directly is that we don't have to maintain any type of data transformation between the API response and the object we are exposing; and the only versioning required in this scenario is the API response data scheme.

Users will need to do `event.detail.addons` to get access to what you are describing. Ideally, that `detail` could be renamed to be `readthedocs` but I don't think that's possible.

Does it make sense to you?

agjohnson: Oh sorry, perhaps we're talking about two separate things. What you're describing for the data structure makes perfect sense. I'd probably still suggest something more descriptive for the attribute name, but the structure makes sense otherwise.

I was trying to describe user programmatic access to the library API, such as:

```js
document.addEventListener("readthedocsdataready", function(event) {
  const readthedocs = event.readthedocs;
  const isLatestVersion = readthedocs.utils.isLatestVersion(event.data.version.current);
});
```

This is a bad example as the logic already exists as data response, but I'm trying to instead just show how the library might be used more directly from the Event. This avoids exposing our library as a global.

There is probably a better naming pattern, this is just an illustration.

humitos: Oh, I see. This makes sense to me.

I haven't thought about this use case yet, but it makes total sense. I'll need to think how we are going to expose our logic for others to be reused.

Thanks for pointing this out. It's useful 👍

agjohnson: Perhaps this can come later too. I thought the intent with `window.readthedocs` was exposing the API to the user, not just the API data structure.

Just considering the API data structure, this PR is pretty close.

Access to the API could just as easily come from an import: `import * from "readthedocs-addons"`

pradyunsg:
Haven't reviewed the logic/code but the description sounds great!

src/readthedocs-config.js
humitos: >  I thought the intent with window.readthedocs was exposing the API to the user, not just the API data structure.

Ah, OK. No, I was not thinking about exposing the API just yet.

> Just considering the API data structure, this PR is pretty close.

👍🏼

humitos: I don't think we will expose our library/data globally and only expose the raw data (without any manipulation) from a custom event. This will reduce the maintenance complexity from our side and give users the exact same data we use to render our addons.

Logic and addons customization will come in a future iteration. I want to allow users to have access to the raw data for now. After that, we can analyze if it makes sense to give them more tools/library/framework to facilitate these customizations in case we need them.

humitos: Oh, by the way, the `event.detail` attribute cannot be renamed. This is just how javascript passes data via a custom event: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent

src/readthedocs-config.js
```diff
@@ -13,17 +28,66 @@ export function getReadTheDocsConfig() {
...
+        console.debug("Error parsing configuration data");
+        return undefined;
+      }
+      return response.json();
+    })
+    .then((data) => {
+      // We force the user to define the `<meta>` tag to be able to use Read the Docs data directly.
+      // This is to keep forward/backward compatibility without breaking integrations.
+      const metadataAddonsAPIVersion = _getMetadataAddonsAPIVersion();
+      if (metadataAddonsAPIVersion !== undefined) {
```
humitos, line 49: Is there a good way to communicate to the developer that they are subscribed to the event but the `<meta>` tag, which is mandatory, is not defined?

I'd like to read in the console:

```
Subscribing to `readthedocs-addons-data-ready` requires defining the <meta ...> tag in the HTML
```

or similar.

agjohnson (commented):
I noted how we could improve the promise flow here, the most important point is probably to be ensuring we stop the promise when we hit an error state though.

src/readthedocs-config.js
```diff
@@ -15,16 +30,66 @@ export function getReadTheDocsConfig() {
...
+          fetch(url, {
+            method: "GET",
+          })
+            .then((response) => {
+              if (!response.ok) {
+                console.debug("Error parsing configuration data");
+                return undefined;
+              }
+              return response.json();
+            })
```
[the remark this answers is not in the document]
humitos: I implemented this suggestion in e8ec7583b363de36bc44d6be8c59fef24a40c37d

src/readthedocs-config.js
humitos: I opened https://github.com/readthedocs/addons/issues/279 to track this and find a way to do it in case it's possible.

humitos:
I updated this PR with the latest changes on `main` and made some minor adjustments. @agjohnson can you please review it?

I opened a POC on the CPython repository that requires the work from this PR to render the language and version selector on their documentation: https://github.com/python/cpython/pull/116966

I've been talking to them via Discord and they are thinking about merging that PR soon to continue with the experimentation about hosting their docs on Read the Docs.

agjohnson (approved):
This looks good to me. Overall, I would still point you towards using Promise instances more often, as this pattern does help tame promise chaining and nesting rather well.

src/readthedocs-config.js
```diff
@@ -15,16 +30,66 @@ export function getReadTheDocsConfig() {
...
+            });
+        } else {
+          const readthedocsDataReady = new CustomEvent(
+            EVENT_READTHEDOCS_ADDONS_DATA_READY,
+            {
+              detail: data,
+            }
+          );
+          document.dispatchEvent(readthedocsDataReady);
+        }
```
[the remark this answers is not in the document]
humitos: I opened https://github.com/readthedocs/addons/pull/286 with this implementation. Please, let me know if this is the direction we want to follow here.

I read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises, but I'm still not 100% sure I'm following the pattern properly.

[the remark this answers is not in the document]
agjohnson: Sorry, I should have clarified in my review, I didn't mean to hold up this PR with reworking this logic. `Promise` is indeed a bit harder to get used to than chaining (or mixing with chaining too). I'll take another look here.

agjohnson (approved):
The promise logic change looks good :+1: 

Edit: oookay, I was confused why the logic didn't seem different here. I saw the commit eb550b above and thought it was in this PR :sweat_smile: 

I just realized this from a commit in #286 instead. I think we can merge this and tune the logic afterwards.

src/readthedocs-config.js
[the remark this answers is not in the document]
humitos: Don't worry. It was clear to me that this chunk of work was not required on this PR --that's why I included in a different one based on this one, just in case it got too complex 😅

humitos:
🎉

PR #66, opened 2023-05-02
Flyout: initial work for the flyout

### LINKED ISSUE #70: Flyout: convert current flyout to web component
As the first step, we want to copy the current flyout code and and convert it into a web component. It should work in the exact same way as the one generated by the current JS code and HTML returned by the endpoint.

It has to be floating at the bottom right for the first iteration.

This will allow us to start slowly migrating new projects to use the new flyout, and expose the other addons.

Front logo [Front conversations](https://app.frontapp.com/open/top_beu13)

humitos:
I'm working on some mockups for the flyout to use the API more and understanding what are the needings for this addon. Now, I'm displaying more or less the same info we used to show in the flyout. However, I think "the new flyout" will end up being the "tool bar/dashboard" we have been talking about in #53 

Screenshot_2023-05-01_18-48-08

Closes #70

### DIFF
--- a/public/_/readthedocs-config.json
+++ b/public/_/readthedocs-config.json
@@ -64,7 +64,12 @@
     },
     "flyout": {
       "enabled": true,
-      "translations": [],
+      "translations": [
+        {
+          "slug": "es",
+          "url": "/es/"
+        }
+      ],
       "versions": [
         {
           "slug": "stable",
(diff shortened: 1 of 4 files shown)


zanderle:
Closing because the work is continued in https://github.com/readthedocs/addons/pull/86

PR #73, opened 2023-06-29
Switch to web-test-runner test suite

agjohnson:
This switches out Jest for web-test-runner, which is the test suite that
is recommended for web component building, especially from Lit:
https://lit.dev/docs/tools/testing/

This replaces manual DOM testing manipulation and instead tests inside
the browser environment instead. This is handy for picking up on
exceptions that wouldn't be caught outside the browser -- I picked up on
a couple just adding notification tests.

The Lit starter kit has some more examples of tests:
https://github.com/lit/lit-element-starter-js/blob/main/test/my-element_test.js

I just added some to test the functionality. This still requires some
mocking, as all imports do need to be JS modules -- CSS and JSON
included in that.

I've wanted to test this for the dashboard, but my environment was broken all
day today, so I started testing here instead. We don't have any tests written
yet anyways, so this is an easier place to start already, not to mention the
footprint is smaller.


agjohnson:
After merging with main (my main was off it seems), I get an error from the semver module:

```
SyntaxError: The requested module './../node_modules/semver/functions/coerce.js?wds-import-map=0' does not provide an export named 'default'
```

There was an explanation for modules that are commonjs, but I'll leave this be for now. I'll wrap this up if there is merit to this proof.

agjohnson:
The issue went a little deep. I was able to get working tests on ext-theme using the Rollup commonjs plugin (recommended by WTR docs). But even with this here, semver 7.x seems to have cyclical import errors going back a few years at very least. None of the fixes mentioned in https://github.com/rollup/plugins/issues/879 and related issues helped.

We're only using the two functions here, so I don't know if we really need to be on the latest release, but 6.3.1 was released this week either way, so it's not ancient by any means.

Otherwise, mocking the static files we're importing directly into the Webpack bundle seemed to solve the remaining errors on tests.

package.json
```diff
@@ -31,7 +32,7 @@
     "jquery": "^3.6.4",
     "lit": "^2.7.1",
     "prettier": "^2.8.4",
-    "semver": "^7.3.8",
```
humitos, line 34: Is there any reason why you downgraded this one?

src/notification.js
```diff
@@ -1,5 +1,4 @@
-import semverMaxSatisfying from "semver/ranges/max-satisfying";
-import semverCoerce from "semver/functions/coerce";
+import * as semver from "semver";
 import { library, icon } from "@fortawesome/fontawesome-svg-core";
```
humitos, line 2: I understand this not a good pattern because by importing everything the resulting "compiled" js contains all the module instead of only the used/required functions making the file size bigger. Am I correct?

package.json
humitos: > But even with this here, semver 7.x seems to have cyclical import errors going back a few years at very least. None of the fixes mentioned in https://github.com/rollup/plugins/issues/879 and related issues helped.

Found the answer 🤷🏼

src/notification.js
```diff
@@ -178,6 +182,8 @@ export class NotificationElement extends LitElement {
   }
 }
 
+customElements.define("readthedocs-notification", NotificationElement);
```
humitos, line 185: Why did you moved this _outside_ the the `NotificationAddon` class?

tests/index.test.js
```diff
@@ -1,15 +1,11 @@
...
-
-test("setup() function is defined", () => {
-    expect(readthedocs.setup).toBeDefined();
+describe("Main library", () => {
+  it("is defined", () => {
+    expect(readthedocs).to.exist;
+  });
+  it("setup() function is defined", () => {
+    expect(readthedocs.setup).to.exist;
+  });
```
humitos, line 10: I'm not sure to understand this, but I suppose I will need to read the documentation to learn how to use this framework :)

src/notification.js
```diff
@@ -72,25 +72,29 @@ export class NotificationElement extends LitElement {
...
+        return this.renderNonLatestVersionWarning();
       }
-    } else if (
-      !IS_PRODUCTION &&
-      this.config.addons.non_latest_version_warning.enabled &&
-      this.highest_version
-    ) {
-      return this.renderNonLatestVersionWarning();
+    } catch (TypeError) {
+      // Configuration options are missing or configuration is empty.
```
humitos, line 88: Why this try/catch is required here?

humitos (approved):
Looks good, but I have some questions to understand why some changes are required.

package.json
agjohnson: Yeah, though there might be a later version that doesn't have the cyclical import issues, or perhaps I just don't understand the bug enough. I took a couple quick swings at this, but none of the suggestions seemed to do anything.

src/notification.js
agjohnson: It follows the patterns used by Elements in the Lit docs and other Elements in this library. When using Typescript, the customElement is a decorator on the class and always is registered.

This helps with testing and allows the Element to be tested directly, instead of having to pass configuration to the Addon just to test the Element class. Given we already have one conditional layer, the addons configuration, the second layer of conditionally registering the Element shouldn't be necessary.

tests/index.test.js
agjohnson: Several JS libraries use similar patterns. This is Mocha, which looks a little like Jest out of the box. The function chaining is an odd pattern to get used to, these are all documented constructs at least though. Each `it()` is a single test case, where `describe()` is a grouping of test cases

src/notification.js
agjohnson: We could tune this down, though I don't know if it will make too much of a difference as Webpack performs tree shaking on the built bundles. It could be bringing in more than we're asking for though.

src/notification.js
agjohnson: This we'll follow up on in #122

agjohnson:
Okay, so the only thing I haven't figured out here is how to get the docdiff/visual-dom-diff/diff-match-patch working with web-test-runner. The error I'm getting is:

```
TypeError: diff_match_patch_1.diff_match_patch is not a constructor
```

This is caused ultimately by rollup/web-test-runner and Rollup's commonjs resolver. The package diff-match-patch bundle does look a little odd, the export is a circular reference object. But this can also be resolved by changing visual-dom-diff to call `diff_match_patch_1()` instead of `diff_match_patch_1.diff_match_patch()`, which is the circular part of the export.

For example, this is what it looks like logging the exports from the package:

```
      {
        __moduleExports: {
          diff_match_patch: [Function: diff_match_patch],
          DIFF_DELETE: -1,
          DIFF_INSERT: 1,
          DIFF_EQUAL: 0
        },
        default: {
          diff_match_patch: [Function: diff_match_patch],
          DIFF_DELETE: -1,
          DIFF_INSERT: 1,
          DIFF_EQUAL: 0
        }
      }
```

agjohnson:
I believe this is actually a Typescript error from visual-dom-diff. I tried a few fixes this afternoon at:

https://github.com/Teamwork/visual-dom-diff/compare/master...agjohnson:visual-dom-diff:agj/es6?expand=1

No luck so far though.

agjohnson:
A couple notes from a call today:

- We're aiming for a mild scope for our test cases, as historically our test coverage was 0%. We might not be able to maintain a very large test suite here, so initially we'll focus on the most important parts first.

- There is a strange behavior when loading the configuration into an Element. That is `(new FlyoutElement()).config = {};`. This seems to trigger the logic at the `index.js` and `getReadTheDocsConfig()` call.

- The remaining tests with visual-dom-diff errors might need a patched visual-dom-diff, but we might be moving in that direction either way.

- Configuration handling is going to be important to writing more tests. We can at least start with loading the fixture file in test cases, ensuring that the mock API response is always the same. We still probably want to be more defensive around our use of the config object however.

One solution is the `object-path` package, which gives us a way to avoid enforcement of the configuration/API schema but does give us a way to avoid missing attributes on the configuration objects:

```js
// For mandatory attributes
if (objectPath.get(this.config, "addons.flyout.enabled", false) !== true) {
  return nothing;
}
// For optional attributes
const downloads = objectPath.get(this.config, "addons.flyout.downloads", []);
if (downloads.length === 0) {
  return nothing;
}
```

I'll try to resolve these last few issues, we'll plan to merge this and try to experiment more with test cases.

cc @zanderle in case you have any experience with web-test-runner or the testing patterns here and have any additional input.

package.json
humitos: We are not requiring this dependency anymore.

src/notification.js
humitos: After our talk, this makes sense to me. We need to _define_ the custom element even if we are not instantiating our addon.

src/analytics.js
```diff
@@ -86,6 +86,6 @@ export class AnalyticsAddon extends AddonBase {
   }
 
   static isEnabled(config) {
-    return config.addons && config.addons.analytics.enabled;
+    return config.addons && config.addons.analytics.enabled === true;
```
humitos, line 89: I think this method is a good place to use the logic for _mandatory attributes_ with `object-path` that we talked. We can check them all at this place and return `false` if we don't have all the mandatory attributes we already know this addons needs to work.

Then, _optional attributes_ could be checked when they are needed. Maybe, it's at render time, or inside a logic that behaves different if they are present.

```javascript
this.enabled = objectPath.get(config, "addons.analytics.enabled", false)
return this.enabled
```

src/flyout.js
```diff
@@ -198,7 +198,7 @@ export class FlyoutElement extends LitElement {
 
   render() {
     // The element doesn't yet have our config, don't render it.
-    if (!this.config) {
+    if (!FlyoutAddon.isEnabled(this.config)) {
```
humitos, line 201: If we move the logic for mandatory attributes to `isEnabled`, I think we don't need this extra check here. We should probably just do

```javascript
if (!this.enabled) {
  return nothing;
}
```

The attribute `this.enabled` is set by `.isEnabled` method that performs all the `objectPath` checks for mandatory attributes.

src/notification.js
```diff
@@ -11,6 +10,10 @@ import { html, nothing, render, LitElement } from "lit";
 import styleSheet from "./notification.css";
 import { AddonBase } from "./utils";
 
+// IS_PRODUCTION comes from Webpack and is undeclared otherwise
+const is_production =
+  typeof IS_PRODUCTION === "undefined" ? false : IS_PRODUCTION;
```
humitos, line 15: This could probably be moved to `src/utils.js` where we have other things like these defined.

src/notification.js
```diff
@@ -67,30 +70,34 @@ export class NotificationElement extends LitElement {
...
     return nothing;
   }
 
   calculateHighestVersion() {
     // Convert versions like `v1` into `1.0.0` to be able to compare them
     const versions = this.config.addons.non_latest_version_warning.versions;
-    const coercedVersions = versions.map((v) => semverCoerce(v));
-    const coercedHighest = semverMaxSatisfying(coercedVersions, ">=0.0.0");
+    const coercedVersions = versions.map((v) => semver.coerce(v));
+    const coercedHighest = semver.maxSatisfying(coercedVersions, ">=0.0.0");
```
humitos, line 100: We are not using this anymore. The PR should be rebased/merged with `main`.

humitos (commented):
This looks good to me. After our talk I have a better understanding about how it works. I'm happy to continue working on this from where you will leave after resolving some of the issues you want to solve.

You can probably make some adjustment to leave this into a mergeable state and I can probably continue from there, so you don't get too distracted about this. It seems we at least discussed some of the patterns we want to implement.

package.json
agjohnson: Indeed, I'll rebase and remove this as this as well.

src/flyout.js
agjohnson: Yeah, absolutely, I think that is on the correct path forward. For now, this logic will be a bit extraneous, I don't feel it's the correct call and is only temporary. Actually, I'll add a note so we don't forget too.

src/notification.js
agjohnson: Good call, I'll move this there

agjohnson:
The answer here was easier than expected! I figured I just needed spend some more time with configuration to figure this out. I found the WTR config for `requireReturnsDefault` and `preferred` [^1]

> "preferred": If a module has a default export, requiring that module always returns the default export, no matter whether additional named exports exist. This is similar to how previous versions of this plugin worked.

So, there is no need for a fork of visual-dom-diff or diff-match-patch anymore, these both import correctly now.

I did not address the 404 on the addons config yet, I might end up doing that in a follow up PR.

[^1]: https://github.com/rollup/plugins/tree/master/packages/commonjs#requirereturnsdefault

agjohnson:
Everything is passing! And without any hacks!

I'd say lets merge this before I get tired of rebasing and we can deal with the follow up work next:

- [ ] Try a test case the alters the Element.config again, it should not request the addons API response from localhost. Instead it should always use the config as if `Element.config = config`
- [ ] Experiment more with test cases and writing some basic tests

I *really* like the `npm run test:dev` command (`web-test-runner --watch`). Be sure to give that a try!

humitos:
Cool! I think I can continue from here. Thanks for the initial setup for all of this 💪🏼 

I will read some docs about testing, work on https://github.com/readthedocs/addons/issues/122, figure it out a pattern to pass the config and experiment on how to write more test cases.

I'm merging this PR since looks ready to me.

agjohnson:
Nice, I'm excited to get this out and start writing some real tests. I also started this switch to web-test-runner on the new dashboard, and I think I can reuse what I learned here to make the change there too.

zanderle:
> cc @zanderle in case you have any experience with web-test-runner or the testing patterns here and have any additional input.

I don't have any experience with web-test-runner, but I'll write my other comments below. Nothing major though, things are looking good to me.
In my experience, I prefer writing more e2e kind of tests, but I don't think that applies here so well.

> A couple notes from a call today:
> 
> * We're aiming for a mild scope for our test cases, as historically our test coverage was 0%. We might not be able to maintain a very large test suite here, so initially we'll focus on the most important parts first.

I think that's a good call.

> * Configuration handling is going to be important to writing more tests. We can at least start with loading the fixture file in test cases, ensuring that the mock API response is always the same. We still probably want to be more defensive around our use of the config object however.
> 

I agree that that would be a good idea.

> One solution is the `object-path` package, which gives us a way to avoid enforcement of the configuration/API schema but does give us a way to avoid missing attributes on the configuration objects:
> 
> ```js
> // For mandatory attributes
> if (objectPath.get(this.config, "addons.flyout.enabled", false) !== true) {
>   return nothing;
> }
> // For optional attributes
> const downloads = objectPath.get(this.config, "addons.flyout.downloads", []);
> if (downloads.length === 0) {
>   return nothing;
> }
> ```

An alternative (if we need it) to this would be to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility).

agjohnson:
> An alternative (if we need it) to this would be to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility).

I'm excited for this syntax, it's quite nice. It's still a fairly new feature of ECMA2020 (?) though, and I never know when to start adopting new syntax anymore. I guess it could be safe given Babel should transform it either way, but at some point we are going to drop Babel too.

humitos:
Optional chaining looks good, yeah. It looks pretty similar to `object-path` but it's built-in. The only thing that I didn't see is the ability to return a default value. If that's not possible the code will have multiple `if (field === undefined)` each time we check for them -- but probably similar to `object-path`, tho.

agjohnson:
Aye, optional chaining doesn't do defaults like object-path, that would be more native syntax, like:

```js
> let config = {};
undefined
> config?.addons?.flyout?.enabled || false;
false
> config = {addons:{flyout:{enabled: true}}};
{ addons: { flyout: { enabled: true } } }
> config?.addons?.flyout?.enabled || false;
true
```

This is nice syntax in most cases, maybe a little less readable than object-path when working with multiple lines of conditionals, like:

https://github.com/readthedocs/addons/blob/2cc9b0d474b8be9cd369893327a2667f839ee32c/src/notification.js#L262-L266

It can be hard to decipher code that jumps between using `||` for comparison OR, and using `||` to provide a default on undefined.

Babel does indeed transform this though, so the feature would be safe to use and probably still wouldn't stop us from dropping es5/Babel soon even. If we start with object-path, we can fairly easily change the code later too.

I might lean towards object-path still, but it probably doesn't make too much difference either. If optional chaining feels right, it should be fine to use.

PR #74, opened 2023-07-12
Rename library to use "addons" everywhere

### LINKED ISSUE #4: Rename library
Before we start using this on user projects/publicly, `readthedocs-client`
sounds like a CLI or API wrapper, and we've called this the doc embed JS so
far. There's probably wisdom is reusing that naming or something similar.

humitos:
Closes #4


agjohnson:
Looks like there are some failures at CI, and this needs formatting and a test fix. I also opened up #73, which has it's own quirks, but would replace the Jest test suite with web-test-runner and Mocha (Mocha is quite similar to Jest in application)

humitos:
Cool. I ran the format and I will merge it after that. I don't fully understand how tests are working and it seems they are going to be replaced anyways with your PR. We can come back to this once we have the suite established.

PR #77, opened 2023-07-12
EthicalAds: create the addon class

### LINKED ISSUE #67: Ads: return `publisher`, `tags` and more in the endpoint
There are some data that comes from `/sustainability/` (-ext) about the publisher, if it's ad free or not and more that should be considered at the moment the sponsorship addon is rendered at https://github.com/readthedocs/readthedocs-client/blob/main/src/sponsorship.js#L12

I think it makes sense to return all this data in the `/readthedocs-config/` endpoint under `addons.ethicalads` key.

Reference API: https://github.com/readthedocs/readthedocs-ext/blob/4fe6331fac357a44f5e1be3879d7475ce560ffe6/readthedocsext/donate/restapi/views.py#L52

humitos:
Show an ad using EthicalAd as a backend and the data coming from `/_/readthedocs-config/` inside `addons.ethicalads`.

Peek 2023-07-12 19-06

Closes #67
Requires: https://github.com/readthedocs/readthedocs.org/pull/10534


src/index.js
```diff
@@ -2,6 +2,7 @@ import { getReadTheDocsConfig } from "./readthedocs-config";
 import * as notification from "./notification";
 import * as analytics from "./analytics";
 import * as search from "./search";
+import * as sponsorship from "./sponsorship";
```
agjohnson, line 5: Any reason not to call this "ethicalads"? That would align with the config key at least, and I don't think we are going to do anything else with _sponsorships_ here otherwise.

src/sponsorship.js
```diff
@@ -1,42 +1,83 @@
...
+ * Sponsoship addon
+ *
+ * Show an Ad in the documentation page.
+ *
+ * Read more at:
+ *  - https://docs.readthedocs.io/en/stable/advertising/ethical-advertising.html
+ *
+ * @param {Object} config - Addon configuration object
+ */
+export class SponsorshipAddon extends AddonBase {
```
agjohnson, line 18: Same question here on naming, but otherwise:

```suggestion
/**
 * Sponsorship addon
 *
 * Show an ad in the documentation page.
 *
 * Read more at:
 *  - https://docs.readthedocs.io/en/stable/advertising/ethical-advertising.html
 *
 * @param {Object} config - Addon configuration object
 */
export class SponsorshipAddon extends AddonBase {
```

agjohnson (approved):
Looks good, my only input right now is on the naming. Only partially related, but in the future I think we should drop sponsorship/donate patterns from the application and just have ads and subscriptions.

humitos:
I updated the hardcoded JSON to return `addons.ethicalads` as well. Now, the ad is shown in our testing page as well 🚀

humitos:
I'm merging this PR. We will probably come back to it once we make the fixedfooter placement to work properly.

PR #79, opened 2023-07-13
EthicalAds: use text-only and fixedfooter

humitos:
Currently, we are going to show always a text-only ad because our initial flyout will be floating at the bottom right. So, we can't show an image at the same placement because they will collide.

A future idea is to use:

- Sphinx + Read the Docs theme:

  Eventually, our theme will integrate the flyout in the sidebar using the new API. At that point, we will be able to use `image` and `stickybox` on this case.

- Any other doctool:

  We will show a text-only ad with the fixedfooter style. This should not collide with nothing in most of the themes/doctools.

### Example

Screenshot_2023-07-13_20-46-58


davidfischer (commented):
Looks good from the ads side and I'll aim to increase the fixedfooter z-index so it also works well in Sphinx.

agjohnson (approved):
Looks good!

PR #81, opened 2023-07-19
EthicalAds: use it as UMD library

### LINKED ISSUE #6: Correct import for ethical-ad-client
This is also an issue with the ea-client, just noting here to find a solution.

The fix for the import statement is that the ea-client package.json is wrong.
The `main` configuration option points to a file that doesn't exist.

`main: "dist/ethicalads.js"` at the ethical-ad-client package configuration solves the import problems here,
and `import "ethical-ad-client"` worked with this change.

humitos:
Install all the requirements and make required adjustments to be able to use `ethical-ad-client` as a js library.

See https://github.com/readthedocs/ethical-ad-client/pull/155
Closes #6


src/ethicalads.js
```diff
@@ -72,9 +72,7 @@ export class EthicalAdsAddon extends AddonBase {
 
   injectEthicalAds() {
     this.createAdPlacement();
-    // FIXME: the function `ethicalads.load_placements()` is called automatically when the module is imported,
-    // but we want to call it manually because we need to inject our `div` first.
-    load_placements();
+    ethicalads.load();
```
agjohnson, line 75: This is technically going to execute the placement loading twice I believe. My PR on the EA client only changed the bundle library to include exports. The client will still be automatically executing the placement loading until we add some conditional workaround to the automatic execution.

PR #86, opened 2023-08-02
Recreating flyout as a webcomponent

zanderle:
Building on top of the work started in https://github.com/readthedocs/addons/pull/66

Closes https://github.com/readthedocs/addons/issues/70


humitos:
I did a quick QA locally and it's looking great! 💯 . I'm super excited about this work and I'd like to deploy it soon and start testing it on production 👍🏼 . @zanderle what's the status of it? Is it ready for review? If not, what are you considering it's missing currently?

zanderle:
@humitos yeah, the first version is ready. One thing I am unsure about is the fact that we're grabbing the versions from `config.addons.flyout.versions` but we're showing the current version from `config.versions.current`. This might be ok, depending on how the endpoint for this will work, but we want to make sure we have a single source of truth here. 

Additionally, this is important so we can mark the current version in the versions list. We could compare with `config.versions.current` or add a `current` flag in one of the versions in `config.addons.flyout.versions`.

Other than that, this is still a v1 of the flyout. I'm wondering if it's ok that the input to the webcomponent is the "global" config (with all the addons, and everything", or if it should be just the `config.addons.flyout`?

humitos:
@zanderle good points raised here!

Originally, I thought that the whole config will be the input for each of the addons. This is because "the current version" for example, is something that more than one addon will require. So, I moved those "shared data" into an upper level and keep the data that is _only required_ for the particular addon under `config.addons.*`.

Besdies, the `config.addons.flyout.versions` may be something different than `config.addons.search.versions` because they require different attributes or the whole filter for the versions is different. However, the `confg.versions.current` is always the same for all the addons.

humitos (approved):
This is great! I'm love with this work 😍 

I tested it locally with different doctools and it works great!

Peek 2023-08-28 15-34

humitos:
I added some small commits with simple tweaks since I want to deploy this tomorrow, together with https://github.com/readthedocs/readthedocs.org/pull/10650

If there are extra changes here, we can open new PRs.

humitos:
I'm gonna merge this since I want to deploy it today to start testing it.

PR #94, opened 2023-08-30
packages.json: update EthicalAd client

humitos:
This version has the fix for z-index.
See https://github.com/readthedocs/ethical-ad-client/pull/167


humitos:
I don't think there is too much to review here 😄

agjohnson (approved):

PR #99, opened 2023-08-30
Enable search + flyout and don't break

### LINKED ISSUE #91: Enabling search breaks flyout
It seems there is a race-condition or similar between the addons. Uncommenting search, makes flyout to not be shown. However, putting the flyout initialization _after_ the search one, makes the flyout to be shown but the search doesn't work.

Even, uncommenting the search addon, makes the EthicalAd one to not be displayed.

I'm pretty confused here 🤷🏼

humitos:
We had a problem when enabling multiple addons because the manipulation of the DOM in an async way. This commit uses a different pattern that solves this issue.

Closes #91

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -29,7 +29,7 @@ export function setup() {
           // Addons that are only available on development
           addons.push(ethicalads.EthicalAdsAddon);
           // NOTE: Disabling search for now because it's incompatible with Flyout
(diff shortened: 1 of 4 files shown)


agjohnson (approved):
This seems like a better way of injecting the additional web component all around :+1:

PR #102, opened 2023-08-31
Enable addons by default

### LINKED ISSUE #69: Search: make the addon enabled by default
Currently, we are only exposing the notifications and analytics addons. The next step would be to expose search as well. However, it will be only discoverable by using the default hotkey (`/`). Note that it won't be integrated with any element in the HTML. Sphinx search will continue working as locally.

In a next iteration, when we have the flyout already working, we will probably expose it via the search text input that lives inside the flyout.

### LINKED ISSUE #76: Ads: make the addon enabled by default
We want to show EthicalAds on documentation pages that are using the new addons js library. For this issue, we need to consider _where to show_ the ad since the flyout will be floating at the bottom right, which is the same placement by default of the ad.

We talked about using text-only fixed footer ads for this purpose as the first implementation.

humitos:
Closes #69
Closes #76


humitos:
This PR just enables all the addons we currently have developed. I'm merging this since I'm preparing everything to deploy next week and I want to start testing this locally more.

PR #104, opened 2023-09-04
EthicalAds: un-bundle the client

### LINKED ISSUE #6: Correct import for ethical-ad-client

humitos:
- remove dependency
- don't require UDM library changes/complexity
- load the dependency only if EthicalAd is enabled for the project (useful for commercial 👍🏼 )
- protect us from blocking `readthedocs-addons.js` by AdBlocker

I decided to go this way because it simplifies things a lot and also it uses the
standard/supported way of using EthicaAd client.

* Closes #6
* Closes #81
* Closes readthedocs/ethical-ad-client#155


humitos:
I'll merge this PR because this is going out today. However, I'm happy to keep the discussion open and revert if we don't think this is a good pattern.

agjohnson:
It's okay for now, but we gain benefits from using packaging here so should make sure to come back and revert this. It's not quite clear what the reason for this PR is, but if there are problems problems with an approach that uses the EA client as a UMD, we should note that so we can these issues when we revert this.

We want packaging versioning here for dependencies to ensure changes to the EA client do not affect older versions of the addons client. Obviously not an immediate concern as addons doesn't really have versions yet though.

agjohnson:
> I mention 4 reasons in the description of the PR about this

I read those, but none of these really seemed to be driving this decision. What I was asking for is why we can't continue with the plan to switch EA to avoid auto-injection -- is there something blocking us from this?

Even if there isn't, the technical changes you point out all seem avoidable while retaining standard NPM packaging patterns. This seems like a good middle ground, so might be my choice at this point.

The middle ground between changing the EA library and this PR is:

- Merge the EA UMD PR. This should be a non-event, the only change is the library bundle is now able to import.
- Revert the package installaiton changes here and continue installing the EA library via NPM standard packaging
- Treat the EA library as a dynamic import chunk. File size optimization is possible with dynamic imports, but isn't a top concern yet. See https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import for the technical background
- Use an approach similar to this PR (but using Webpack dynamic imports) to only load the EA library only when needed and after we've injected our placements. See ext-theme for an example of what this looks like in code: https://github.com/readthedocs/ext-theme/blob/main/src/js/build/detail.js#L154-L178 

This doesn't require changing the EA client besides the UMD output option but still accomplishes all of the original goals we had for the ads web component. It would look similar to this PR, just with an internal library instead.

@humitos I'm not sure of the context around protecting us from blocking though. Can you expand on what you mean here? Did we hit a problem somewhere with our JS being blocked?

> If we want to pin the EA client, we can still use this approach and use the URL that has the version on it -- so we are not always on the latest one.

Packaging doesn't seem to be the underlying problem here though. I would save this as a last resort, but don't think we need to recreate packaging in this way to obtain pattern that works for the EA library.

humitos:
> @humitos I'm not sure of the context around protecting us from blocking though. Can you expand on what you mean here? Did we hit a problem somewhere with our JS being blocked?

`ethicalads.io` is a domain that it's blocked by many popular AdBlockers (e.g. EasyList -- see https://easylist.to/easylist/easylist.txt). We also wrote a blog post about this some time ago (https://www.ethicalads.io/blog/2018/05/ads-and-ad-blockers/).

If we embed the Ethical Ad client _inside_ our `readthedocs-addons.js` I suspect they could block this file due that it contains the Ethical Ad client and _remove all the addons completely_. Reverting this situation will be pretty hard. I think that loading the client from it's original URL protects us from this scenario. cc @davidfischer

agjohnson:
> ethicalads.io is a domain that it's blocked by many popular AdBlockers

Yeah, I was aware of that, but if there isn't a specific issue that we've hit with having the library included in our bundle, then I'd say we're safer treating the EA library as an import.

If anything, including the EA domain in the addons library might make it *more* apt to be blocked, as the blocked domain is in our source code now. I don't think blockers go this deep though, so likely neither case is going to be a strong reason to change our packaging patterns.

Using Webpack chunks will:

- Reduce the library file size by splitting the EA code to a separate file
- Reduce concerns that adblockers will block *all* of our JS, as the separate file request would isolate EA code.

Neither are huge concerns to me just yet though. To get all of these benefits, we would upload multiple files for each addons library version, so there is some more work there. But we don't have competing packaging patterns at least.

PR #107, opened 2023-09-05
Develop: delete old flyout files

humitos:
These are not used anymore. It's called "new-flyout" currently.


agjohnson (approved):
It's probably also worth renaming "new-flyout"

Issue #90, opened 2023-08-28
Research about how to integrate web components that depend each others


humitos:
We need to find a way to include `readthedocs-search` inside `readthedocs-flyout` so when clicking on the "Search" input field from inside the flyout, the `readthedocs-search` web component (modal) is triggered.

![Screenshot_2023-08-28_15-47-22](https://github.com/readthedocs/addons/assets/244656/efe87117-c9f3-4a2f-a26f-5c7a7456f94a)

humitos:
We found a simple pattern that consists on listening and triggering a `CustomEvent()`. The event is listened by the addon that has to be enabled and triggered from the addon where the click is done by the user. This seems like an easy pattern that we can follow for now.

The situation it's a little different for those addons that include UI elements that are not modals. I'm thinking about `docdiff`, for example, that has a checkbox. However, the approach for this case could be to implement the checkbox directly on the `flyout` and make this checkbox to trigger the custom event to hide/show the diffs. Here, this pattern seems to work still.

I'm sure there would be more complex addons with more UI elements where this pattern may not work great, but I think we've solved this problem for now and we can revisit in the future if we find it's not enough.

PR #108, opened 2023-09-05
Trigger search addon from flyout input

humitos:
This is an initial attempt to be able to render the `readthedocs-search` addon inside the `readthedocs-flyout` one.

I was able to:

- include the `readthedocs-search` tag inside the flyout
- connect the `focusin` event to the input from inside the flyout to call the search

Each time I click on the input from inside the flyout, it shows a log in the console (for debugging). However, due to some reason that I don't know yet, the `readthedocs-search` tag does not contains any HTML inside ito, so even when the modal is triggered, nothing is shown.

Related #90


humitos:
I was able to make this work with a simpler approach. However, I would need some feedback here and maybe some javascript guidances since there some things that can be improved.

I want to read more documentation and play a little more with this pattern, but it seems it's the way to go.

### Working example

Peek 2023-09-05 14-29

agjohnson:
Before we invest too heavily into using shadow DOM internals, I feel there are some easier solutions here that we can explore first:

Is there a reason we *need* to nest the web components in the first place? The search input is a modal that is separate from the flyout, and that users can implement in a doc native search input element. The flyout only requires a *method to make the search modal visible*.

If there is not a strong reason for this nesting, we only need a pattern for making the search web component modal visible from an element in the flyout web component.

The path forward I'm describing is:

- Search web component handles a custom DOM event, for example the [Lit docs on handling an event from another web component](https://lit.dev/docs/v1/components/events/#fire-event) shows how custom events work between two web components.
- The flyout UI element to trigger search would fire this event off, ie `@focus="${this.showSearch}"`

This should remove the need for nesting, but also only requires that the search web component be defined once, globally, on the page.

humitos:
This feedback is great, thanks! I will read a little more and explore your ideas.

humitos:
> Is there a reason we _need_ to nest the web components in the first place? The search input is a modal that is separate from the flyout, and that users can implement in a doc native search input element. The flyout only requires a _method to make the search modal visible_.

Exactly. Originally, I tried to connect the input from inside the flyout to trigger the modal, but I wasn't able to. However, while I was doing the research to nest the search, I found how to connect the event with the input from inside the flyout -- but at that time I forgot that original idea 😅 

_However_, I think this research will be useful for addons like docdiff (or similar ones) that has a UI (small checkbox for the docdiff currently). So, I guess that at some point we will need to nest them, maybe?

> The path forward I'm describing is:

I will check this path 🚀

agjohnson:
> So, I guess that at some point we will need to nest them, maybe?

Good question. Yeah, especially for docdiff that is still unknown. The docdiff web component is a bit unique, in that it has two components really -- the UI slider and the actual DOM manipulation part. I could see authors wanting to customize both, maybe independently even?

humitos:
Man, I implemented your idea and it was pretty simple 🚀 . I think this is ready to get merged.

src/search.js
```diff
@@ -506,24 +506,36 @@ export class SearchElement extends LitElement {
...
       let element = document.querySelector(this.triggerSelector);
-      if (element !== undefined) {
+      if (element !== undefined && element !== null) {
         element.addEventListener(this.triggerEvent, this._handleShowModalUser);
       }
     }
+
+    // The "readthedocs-search-show" event is triggered by "readthedocs-flyout" input
+    document.addEventListener(
+      "readthedocs-search-show",
```
agjohnson, line 516: I could see us wanting to reuse this pattern a number of times perhaps. Maybe it makes sense to have event names defined as constants in a separate, `events.js` import?

agjohnson (approved):
:muscle: This is great! I'm glad this option worked out well for us, hopefully this pattern can work for docdiff and the flyout menu too.

src/search.js
humitos: I moved this to an `event.js` file and import the event from the addons.

PR #114, opened 2023-09-07
Remove check for "compiled assets up-to-date"

humitos:
Do not check if the `dist/` compiled assets are up-to-date on CircleCI. This has been pretty annoying to work with.
Each time you add a commit, you have to re-compile the assets. Then if another PR is merged, you have to update your branch and re-compile the assets.

Now, since we are adding a release process that explicity re-compile the assets before doing the release, compiling them on each PR doesn't make sense anymore.

We will be always using the compiled assets from GitHub tags now when deploying.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,20 +14,11 @@ commands:
       - run: npm ci
       - run: npm run lint
   run-build:
-    description: "Ensure compiled assets are up to date and HTML is valid"
+    description: "Ensure assets compile properly"
     steps:
       - checkout
       - run: npm ci
       - run: npm run build
-      - run:
-          name: Ensure built assets are up to date
-          command: |
-            if [[ `git status dist/ --porcelain` ]]
-            then
-              echo "ERROR: assets are out of date. Make sure to run 'npm run build' on your branch."
-              git status dist/ --porcelain
-              exit 1
-            fi
 
 jobs:
   checks:


agjohnson:
It's annoying, but all the repositories follow this pattern currently because the compiled binaries are included in the repository. If there are binaries in `dist/` in the repo, they should always reflect the code in the repository.

If we can remove the files in `dist/`, then this piece can be removed though. I think we've moved away from this pattern on this repository.

But for instance, I would not make this change on ext-theme even though it suffers from the same annoyances with tracking binary files. We use the binary files directly in the case of ext-theme, sphinx_rtd_theme, etc.

humitos:
> It's annoying, but all the repositories follow this pattern currently because the compiled binaries are included in the repository. If there are binaries in `dist/` in the repo, they should always reflect the code in the repository.

This is _our_ repositories, or it's more like a standard in the javascript world?
 
> If we can remove the files in `dist/`, then this piece can be removed though. I think we've moved away from this pattern on this repository.

We can remove `dist/` from this repository if we want to, actually. There is no reason to have this folder here. However, we will need to compile the bundled files when deploying the addons to S3.

> But for instance, I would not make this change on ext-theme even though it suffers from the same annoyances with tracking binary files. We use the binary files directly in the case of ext-theme, sphinx_rtd_theme, etc.

Regarding the `sphinx_rtd_theme` I understand we talked about removing them from the repository in multiple opportunities but we haven't done it yet because there are people installing it from Git only. I don't know how we are using the bundled files in the ext-theme, tho, so I know nothing there.

agjohnson:
> This is our repositories, or it's more like a standard in the javascript world?

Well, because we aren't using packaging proper, it's a little of both. If you look through repositories for projects on NPM, you won't find these compiled files. However, you will see most NPM packages installed with these files in `node_modules/`. These built files are attached during the packaging process, similar to Python, and so don't need to exist in the repository.

We could consider the usage of proper packaging, AWS has a solution for private package indexes even. But because we're treating the repository as the installation source for our repositories, they need to have the built binaries included.

Otherwise we have to compile these packages at runtime. We used to do this on the production servers with CSS using django-compressor and it made development tedious and hard.

I'd support using proper packaging though. It gets us package versioning at the same time -- for ext, ext-theme, corporate, etc. This is maybe far off and a different conversation.

> We can remove dist/ from this repository if we want to, actually. There is no reason to have this folder here. However, we will need to compile the bundled files when deploying the addons to S3.

This still seems fine here, as I don't think we have a use case for installing from the repo.

humitos:
Gotcha! I think we are not ready yet for that "proper packaging" you are mentioning and I don't want to complicate the deployment of the addons currently. I created a script to easily do this in https://github.com/readthedocs/readthedocs-ops/issues/1400 but it requires the files bundled in the repository.

So, for now, I added pre-commit in https://github.com/readthedocs/addons/pull/116 to check that everything is correct before doing a commit 😄 -- simple enough for now.

agjohnson:
Aaah okay, I thought we were compiling on release. I'm still okay compiling at pull merge time, though agree it makes pull request merging annoying.

I'll move the packaging discussion to a meta discussion. Agreed it doesn't need to be a solution here, but it could be a future improvement to make development, and maybe ops, a little bit easier.

Issue #80, opened 2023-07-17
Hotkeys: find a starndar way to handle them


humitos:
We want to understand how to handle keyboard hotkeys for our addons in a standard and well supported way. Note that we cannot just use `/` to trigger the search or `Ctrl + d` to show the docdiff because those are hotkeys the browser (and the user) expect to behave differently.

Here, we would like to find a way to make this in a friendly way for our users. We need hotkeys for:

- enable/disable docdiff
- show search modal
- etc

_(from https://github.com/readthedocs/addons/issues/78#issuecomment-1634820776)_

humitos:
GitHub uses `?` on most of its pages to show a modal with the keyboard shortcuts:

![Screenshot_2023-07-19_15-31-56](https://github.com/readthedocs/addons/assets/244656/fd0074dc-98aa-45b0-8720-10af14fa93f3)

The documentation for all the keyboard shortcuts is at https://docs.github.com/en/get-started/using-github/keyboard-shortcuts

I also found they have a "Managing accessibility" that mentions this hotkeys can be disabled from a setting: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-accessibility-settings

So, I think we can move forward with the original idea of using the hotkeys we consider and add a way to disable these hotkeys with a UI element and save the config in a _cookie_ (we probably don't want to use user-specific settings because we want to be able to cache `/_/addons/`, see https://github.com/readthedocs/readthedocs.org/issues/10536)

humitos:
I found people talking about `accesskey` HTML attribute, but it seems it's not well supported and its usage is discouraged:

> because browsers and assistive technologies implement accesskey support inconsistently, ineffectively, or not at all, the accesskey attribute should typically be avoided or implemented with great care.
>
> Due to numerous problems with implementation, accesskey is typically best avoided

_(from https://webaim.org/techniques/keyboard/accesskey)_ and more on https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey

zanderle:
I don't have deep experience with hotkeys, but I think it's a good approach to try and copy some of the bigger sites. I think Github is a great example.

[TailwindCSS](https://tailwindcss.com) comes to mind as well with its "ctrl/cmd + K" hotkey for searching. I bring them up because I like how they show that shortcut in the search bar itself:

[...]

Issue #25, opened 2023-03-29
Search: prevent triggering other integrations with `/`


humitos:
It seems that `/`, `?`, `Ctrl + f` and `Ctrl + k` are pretty common hotkeys to trigger modals (eg. search, help, etc). We are currently mapping our search modal to `/`.

MkDocs, for example, uses `/` to show a modal with some help about the hotkeys. This conflicts with our own modal:

![Peek 2023-03-29 23-32](https://user-images.githubusercontent.com/244656/228673488-b5dfcbe5-67fe-41c7-990a-c9a804cc953c.gif)

I tried to find how to call `document.removeEventListener("keydown");` but it seems it requires passing the function handler as second attribute. What would be a possible solution for these cases?

humitos:
Interesting. Material for MkDocs also uses `/` to jump into the focus of the search input. However, it does not collide with ours for some reason 🤷🏼 . I'm testing this locally at http://test-builds.devthedocs.org/en/mkdocs-material/

humitos:
Currently, we are injecting our script at the end of `<head>`. Would be possible to solve this issue by injecting our script at the beginning of it, _making it loading in a sync way_ and using `event.preventDefault()`?

agjohnson:
Just wanted to note that `C-f` is a browser shortcut (find in Firefox and others) that we shouldn't override.

> Would be possible to solve this issue by injecting our script at the beginning of it, making it loading in a sync way

We still want to load async or we'll block the page load, so async might still load *after* other async scripts still.

Noted on #115, but I'd probably say we shouldn't do anything unexpected here and let the project maintainer decide this instead. They can control their own project's behavior on keyboard shortcuts.

Long term, we need UI that will trigger these elements, keyboard shortcuts are more of a temporary solution because we don't have this UI.

humitos:
In https://github.com/readthedocs/addons/pull/115#issuecomment-1710908540 we talked about documenting how to disable conflicting hotkeys from other doctools and point users to this documentation: https://github.com/readthedocs/addons/pull/115#issuecomment-1710908540

FeeeeK:
Apparently this does conflict with mkdocs, you can clearly see it [here](https://vkbottle.readthedocs.io/ru/dev/), this documentation uses a custom build process, maybe this is the reason.
Also, clicking on the search string from mkdocs-material opens the readthedocs search and this is very annoying.

[...]

PR #115, opened 2023-09-07
Focus search input when modal shown

### LINKED ISSUE #78: Search: when hitting `/` it shows the modal but it does not focus it
The focus goes to the default search input from Sphinx. I'm testing this locally in http://test-builds.devthedocs.org/en/manual-integrations/

I think it's because our theme is catching this event as well.

humitos:
This is a small hack for now.
We should find a cleaner way to solve this issue.
It works fow now :)

Closes #78


agjohnson:
Part of me feels like we should honor the site/theme keyboard shortcut, even though it's in conflict with our UI. This is the main problem with relying on keyboard shortcuts, it's always an arms race of sorts to ensure your code captures the event first.

The approach with firing a timer off probably works for most cases, but it will effectively duplicate the key press event. If the theme/site handler for the existing key press has side effects other than focus, like showing up of a separate search modal or something, then the reader will get that UI *and* our search modal UI. That is, the use of a timer doesn't intercept the event, but duplicates it.

I know you can manipulate previously registered event handlers, however I've never done this for third party code. Normally, you know the exact function you are trying to deregister. If you can find the existing handler, it's possible to wrap that handler and intercept the keypress event for just `/`, bubbling all other event through to the handler.

humitos:
> Part of me feels like we should honor the site/theme keyboard shortcut, even though it's in conflict with our UI.

I'm not sure what this means. Can you expand on this?

> The approach with firing a timer off probably works for most cases, but it will effectively duplicate the key press event. If the theme/site handler for the existing key press has side effects other than focus, like showing up of a separate search modal or something, then the reader will get that UI _and_ our search modal UI. That is, the use of a timer doesn't intercept the event, but duplicates it.

Yeah, I've already hit this with Material for MkDocs, yeah. I'm not sure what to do in those cases.

> I know you can manipulate previously registered event handlers, however I've never done this for third party code. Normally, you know the exact function you are trying to deregister. If you can find the existing handler, it's possible to wrap that handler and intercept the keypress event for just `/`, bubbling all other event through to the handler.

I researched about this and I wasn't able to de-register event handlers from others. I'm tracking this case and the one from the previous paragraph in https://github.com/readthedocs/addons/issues/25

----

I know the hack from this PR is not good, but I don't have anything better at the moment and I thought that at least it was better than the current behavior, but I know this is _not_ the solution 🤷🏼

agjohnson:
> I'm not sure what this means. Can you expand on this?

Just that we shouldn't try to replace the override keyboard shortcut. These shortcuts may not be there for any particular reason, but unfortunately the shortcuts might also be important for the site.

> Yeah, I've already hit this with Material for MkDocs, yeah. I'm not sure what to do in those cases.

Ah yes, that has a UI there too. There really isn't a great answer here, as we don't control the website we're injecting the features into. I don't think this is the primary method we want for triggering this UI long term anyways though, so maybe this is okay.

This, among other reasons like screen reader overlap, is why I'm generally -1 on implementing keyboard shortcuts (and I've been known to use qutebrowser!)

> I researched about this and I wasn't able to de-register event handlers from others.

Yeah, I can only say it's possible when you have the instance of the function handy. Otherwise, I don't know how feasible deregistering is.

> I know the hack from this PR is not good, but I don't have anything better at the moment

I might lean towards not doing anything magical or unexpected here. If the user wants our search/flyout/etc, they should disable whatever is interfering with the hotkeys. Long term, we should avoid relying on shortcuts to trigger our UI, but I understand this is just a short term solution to avoid UI that we haven't added yet.

We could make a secondary keyboard shortcut available perhaps? ie `/` or `C-k`?

humitos:
> Ah yes, that has a UI there too. There really isn't a great answer here, as we don't control the website we're injecting the features into. I don't think this is the primary method we want for triggering this UI long term anyways though, so maybe this is okay.

Agreed. I imagine we can probably just document "How to disable the default search on Material for MkDocs" and "How to disable the default search on Sphinx" and other doctools giving people a chunk of code or similar and rely on that.

> This, among other reasons like screen reader overlap, is why I'm generally -1 on implementing keyboard shortcuts (and I've been known to use qutebrowser!)

Regarding using or not using keyboard shortcuts at all, we have this conversation opened: https://github.com/readthedocs/addons/issues/80. I was on the fence, but considering that GitHub makes extensive usage of them, I'm kind of convinced they are not bad.
 
> I might lean towards not doing anything magical or unexpected here. If the user wants our search/flyout/etc, they should disable whatever is interfering with the hotkeys. Long term, we should avoid relying on shortcuts to trigger our UI, but I understand this is just a short term solution to avoid UI that we haven't added yet.

Yes, even having a UI element this may be required. Well, at least right now that we are using an `input` field inside the flyout that triggers the search modal. This is because the focus stays in the `input` field from the flyout instead of jumping to the `input` from the search modal. However, with this PR, it works fine and moves to the `input` from the search modal. At least, it fixes that issue 😄 

> We could make a secondary keyboard shortcut available perhaps? ie `/` or `C-k`?

Yes. I did some tests with `C-k`, but it requires a little more work since you also have to check that none of the other modifiers are pressed and things like that. It was not trivial / quick when I tried it. GitHub uses `s` or `/` to open the search bar (see #80).

humitos:
Are we OK merging this half-baked solution for now? At least it solves the problem in our own theme. We can continue with the proper solution in the other issues that I linked.

agjohnson:
> Agreed. I imagine we can probably just document "How to disable the default search on Material for MkDocs" and "How to disable the default search on Sphinx" and other doctools giving people a chunk of code or similar and rely on that.

:+1: This seems like a great place to start for now. If we do have to hack things in later to support themes/tools, I'd consider that work polish level adjustments for later in our roadmap. I bet we'll find better solutions in the meantime though.

> Regarding using or not using keyboard shortcuts at all, we have this conversation opened: https://github.com/readthedocs/addons/issues/80. I was on the fence, but considering that GitHub makes extensive usage of them, I'm kind of convinced they are not bad.

Aye, though I'm generally -1, I still feel shortcuts are okay as an addition to addons.

I just don't like focusing on this as a primary part of our UX as it's much harder to discover and use as a feature, *especially* for less technical users. As an alternative UX, it's usually okay though, as long as adaptive technologies and normal browser use aren't negatively affected.

> Yes. I did some tests with C-k, but it requires a little more work

Yup, modifiers make this more complicated. Keyboard support is a rabbit hole and there is a lot that goes into doing keyboard support well.

> Are we OK merging this half-baked solution for now?

There are definitely some downsides to this approach, and I feel there is probably a more standard fix for the problem. The focus issue is almost certainly either an async code issue or a matter of CSS transitions throwing the timing off.

I'd try to rule out async troubles first, as it seems the most likely. I believe we need to be using the `updateComplete` event, as the `focus()` call is firing before the template renders. For example, this fix describes `updateComplete` usage on an element in the template:

https://github.com/lit/lit-element/issues/605#issuecomment-471316709

Otherwise, I would have assumed that the Element.updated event is what we would have wanted to sync the event up correctly:

https://lit.dev/docs/v1/components/lifecycle/#updated

I'd give those a try first if you haven't already.

humitos:
> Otherwise, I would have assumed that the Element.updated event is what we would have wanted to sync the event up correctly

Excellent! This was it 🚀 . In fact, in the documentation link you shared, there is the exact same case I need here 😅 . It works perfect now without any type of hack.

> Aye, though I'm generally -1, I still feel shortcuts are okay as an addition to addons.

Man, I don't know if this is what you are thinking here but maybe the "keyboard shortcuts" is an addon on itself that people can enable/disable 🤯 . I'm now convinced _this is the way to go here_. I will open an issue for this.

humitos:
> Man, I don't know if this is what you are thinking here but maybe the "keyboard shortcuts" is an addon on itself that people can enable/disable 🤯 . I'm now convinced _this is the way to go here_. I will open an issue for this.

Boom! 💣 I implemented this idea in https://github.com/readthedocs/addons/pull/118

humitos:
> > Agreed. I imagine we can probably just document "How to disable the default search on Material for MkDocs" and "How to disable the default search on Sphinx" and other doctools giving people a chunk of code or similar and rely on that.
> 
> 👍 This seems like a great place to start for now. If we do have to hack things in later to support themes/tools, I'd consider that work polish level adjustments for later in our roadmap. I bet we'll find better solutions in the meantime though.

Sphinx also catches `/` to focus on the search input from the navbar: https://github.com/sphinx-doc/sphinx/blob/fcc38997f1d9b728bb4ffc64fc362c7763a4ee25/sphinx/themes/basic/static/doctools.js#L146

It seems there is an option to disable it, `DOCUMENTATION_OPTIONS.ENABLE_SEARCH_SHORTCUTS`, but I didn't dig enough to know _where_ to put that yet.

humitos:
This PR is ready to be merged 👍🏼

agjohnson (approved):

agjohnson:
Excellent, I'm glad this approach worked out. I think your idea to split out keyboard shortcuts as it's own addon is an interesting one, definitely worth testing that more. But I like that we have a way to surface configuration of hotkeys already :+1:

PR #116, opened 2023-09-07
Set up pre-commit

humitos:
This is the continuation of the work done in https://github.com/readthedocs/addons/pull/114 since we are not yet ready to remove the `dist/` folder from the repository.

`pre-commit` at least will make this process a little less tedious and automated 😉 . It calls `prettier` and also checks if the `dist/` folder is out of date. In that case, we just need to call `git add dist/` to add the new bundled files compiled by `pre-commit`.

Closes #114

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,24 @@ repos:
     hooks:
       - id: prettier
         args: ["--write", "src/**", "*.js", "*.json"]
+
+  - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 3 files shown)


agjohnson (approved):
This is a great idea :+1: 

We should set up something similar on ext-theme and maybe sphinx_rtd_theme, or `common/` I suppose. Both repos have similar checks in CI and it would be great to catch these before pushing up for review.

humitos:
Cool! I've opened https://github.com/readthedocs/common/issues/189 to track the "sharing among different repos" task.

PR #119, opened 2023-09-09
Search: use the new hotkeys addons to handle hotkeys

humitos:
This show how simple this pattern is 💯 . It uses https://github.com/readthedocs/addons/pull/118 for the hotkeys addons.


src/hotkeys.js
```diff
@@ -30,28 +31,53 @@ export class HotKeysElement extends LitElement {
     this.config = config;
     this.docDiffHotKeyEnabled = this.config.addons.hotkeys.doc_diff.enabled;
     this.docDiffShowed = false;
+
+    this.searchHotKeyEnabled = this.config.addons.hotkeys.search.enabled;
```
agjohnson, line 35: Another try/catch point here.

It could be worth consolidating all of our configuration logic with a wrapper class or something. This would give us a chance to ensure the data structure is always the same and our fields always exist. It's not a thing to solve here, but perhaps worth discussing more.

src/search.js
```diff
@@ -480,61 +483,39 @@ export class SearchElement extends LitElement {
...
-      // prevent opening "Quick Find" in Firefox
-      e.preventDefault();
-      this.showModal();
-    }
+    e.preventDefault();
+    this.closeModal();
   };
 
-  _handleShowModalUser = (e) => {
+  _handleShowModal = (e) => {
```
agjohnson, line 491: This seems to be the same function as above, the name is the same at least. I'm guessing the property above was supposed to be `_handleCloseModal` instead?

agjohnson (approved):
I like this approach, it's much cleaner than individual addons handling their own events.

humitos:
> I like this approach, it's much cleaner than individual addons handling their own events.

I liked how we reach to this approach by "thinking aloud on GitHub and brainstorming these concepts". It was a good team construction 💯

src/hotkeys.js
humitos: Agreed. I opened https://github.com/readthedocs/addons/issues/122 to talk more about this 👍🏼

src/search.js
humitos: Yes. Good catch!

PR #120, opened 2023-09-09
Announcement: initial addon implementation

### LINKED ISSUE #109: Announcement: allow users to publish notes on their docs
I remember talking to someone from the core team about exposing the ability to make "announcements" to project maintainers on their documentation. I'm thinking about something similar to what Webpack documentation is doing.

Screenshot_2023-09-05_21-40-40

The "announcement addons" could be used to publish news or similar in all the pages of the project's documentation. I haven't thought too much about this and the idea could be expanded after some brainstorming, but I think it has potential. Definitely, pretty useful if you want to add these announcements to old versions without rebuilding them. Users should be able to configuration the "versions affected" and also the "message" to show.

Live example: https://webpack.js.org/concepts/

humitos:
It allows users to show a small announcement at the bottom-left of the page. The message and on which versions to show it are configurable.

Screenshot_2023-09-09_18-26-11

Closes #109


agjohnson:
What sort of use case are you picturing for an addon like this?

I get *how* the user might use this addon, but I'm not yet seeing *why* users might want this. This seems like something the user would want to implement without relying on RTD.

There is overlap with the existing notification addon too, which does have similar alternative styles but doesn't yet take custom content. Perhaps this is all the same thing and we want to instead allow users to create custom notification classes?

humitos:
This was an idea about "sharing users' data between the build process and the addons". I use a small implementation of an addon for this that just shows the message in a similar way than Webpack docs does: https://webpack.js.org/concepts/

I imagined a pattern like this:

1. write some text in a form field from the dashboard
2. select on which versions this addon will be enabled

Then, when the addon loads, just fill the announcement with the text coming from the backend (if `announcement.enabled = true` on this version)  _without_ the need of rebuilding the documentation --which is something authors usually want on pretty old versions.

I'm closing the PR for now, but we will come back to this pattern eventually.

PR #125, opened 2023-09-13
Notification: simple stable-latest version warning logic

### LINKED ISSUE #71: Notifications: logic behind non-latest version
It seems there is some hidden logic in the old implementation of this feature. This comment called my attention regarding branches vs. tags: https://github.com/fabric/fabric/issues/2255#issuecomment-1553256375. There is a link to https://github.com/readthedocs/readthedocs.org/issues/3268 which explains a little bit of the history here.

We should clearly define what's the logic behind our semver comparisons.

humitos:
I implemented the simple idea exposed in #71.
I think we can start with this simple logic that's easy to explain and grow from here if necessary.

- on `latest`: a notification to warn the user about some features may not yet be available is shown
- on non-`stable` nor `latest` versions: a notification to warn the user about possible reading and old/outdated version is shown

Note that both notification are pretty simple, it does not require external dependencies (e.g. `semver`) and it's easy to explain to users. The warning is shown only if `stable` is an active version --since we link to it from both.

_However_, I think the best way for users to make usage of this addon is by implementing the logic that fits their needs by themselves. We don't allow users to customize this logic just yet, but we plan to do it in the future when the addons is more settled down.

Closes #71


humitos:
I received some small feedback to "visually differentiate" the notifications. So, I'm using a "bird" icon to denote "canary version" of this documentation when showing the `latest` notification.

astrojuanlu:
I love the ["canary in the coalmine"](https://en.wikipedia.org/wiki/Sentinel_species#Canaries) metaphor but not sure if all users could relate to that 😅 what about something like GNU I.M.P. does with their beta versions, using "experimental" metaphors like  🧪 or 🧑‍🔬 or ☣️ or similar?

image

In any case, I think this is too easily bikesheddable, so just my two cents 🏳️

astrojuanlu:
And maybe for "old" versions something that relates to "time", like ⌛ or 🕰, or even 🥀, or "abandoned" like 🕸

humitos:
I think they look great! 🚀 

Screenshot_2023-09-14_17-50-59


----

Screenshot_2023-09-14_17-50-06

src/notification.js
```diff
@@ -1,10 +1,9 @@
-import semverMaxSatisfying from "semver/ranges/max-satisfying";
-import semverCoerce from "semver/functions/coerce";
 import { library, icon } from "@fortawesome/fontawesome-svg-core";
 import {
   faCircleXmark,
+  faVial,
```
agjohnson, line 4: I also like `flask` and have used that in similar places.

image

https://fontawesome.com/icons/flask?f=classic&s=duotone

agjohnson (approved):
These look great! I agree the differentiation is a solid improvement.

humitos:
I'm merging this PR and it will be deployed tomorrow.

Also, I'm creating two issues to work on the feedback that I received from a conversation on slack: https://readthedocs.slack.com/archives/C029M3W01KR/p1694685981299249

PR #129, opened 2023-09-14
Flyout: Collapse flyout when clicking outside of it

### LINKED ISSUE #110: Flyout: collapse when clicking outside it
I found myself wanting the flyout to collapse automatically when I click on a link and change to another page. I think it also makes sense to collapse the flyout just when clicking _outside_ it while staying in the same page.

Peek 2023-09-05 22-10

zanderle:
Closes #110


agjohnson (approved):

agjohnson:
Looks like this might need rebased. Hopefully my updates in main will resolve all the checks though.

humitos:
I tested this locally and it works great! 🎉

PR #134, opened 2023-09-18
pre-commit: we need types_or for this

humitos:
Run our own hook over CSS or JS or SVG.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,5 +2,5 @@
   name: ensure assets are bundled
   description: Make sure that `npm run build` was ran before commiting.
   language: script
-  types_or: [css, javascript, svg]
+  types_or: [css, javascript, svg, json]
   entry: check-build-assets.sh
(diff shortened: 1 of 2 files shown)


humitos:
I finally figured it out why it was not working properly 🎉

.pre-commit-hooks.yaml
```diff
@@ -2,5 +2,5 @@
   name: ensure assets are bundled
   description: Make sure that `npm run build` was ran before commiting.
   language: script
-  types: [css, javascript, svg]
+  types_or: [css, javascript, svg]
```
agjohnson, line 5: Do we want json as well here? For `package.json` `devDependencies` updates or similar?

agjohnson (approved):

.pre-commit-hooks.yaml
humitos: Yes, good catch!

```suggestion
  types_or: [css, javascript, svg, json]
```

PR #136, opened 2023-09-19
Update check build assets

humitos:
We need to check for the `git diff` because we want those changes to be stashed.

After some local testing, I think this is the _final_ version 😄


humitos:
"git status" was good for CircleCI, but it always returns exit 1 locally because changes are not committed.

agjohnson:
Gotcha, that makes more sense then.

agjohnson (approved):

PR #141, opened 2023-09-19
Remove import of package.json and replace with bumpver control

agjohnson:
Use a string through bumpver instead of dynamic import to get the version
string.


agjohnson:
Whoops, should have based this differently maybe. The only commit here is [5fc8ec0](https://github.com/readthedocs/addons/pull/141/commits/5fc8ec0edb7fb4921e967e9c5c1cbffe0e812333)

The rest are from:

- #73

humitos (approved):
This is good! It needs to be rebased because something weird happened with the history 🤷🏼

agjohnson:
Okay, rebased! Should be cleaner now.

PR #148, opened 2023-09-21
Move HTML attributes outside the constructor

humitos:
The DOM does not support to add HTML elements with attributes. They have to be enabled without attributes and then those attributes have to be added.

Otherwise, the following exception is raised:

```
DOMException: Failed to construct 'CustomElement': The result must not  have
attributes
```

This commit moves the `className` attribute to `loadConfig` which is executed _after_ the element is injected in the DOM.


src/search.js
```diff
@@ -76,8 +73,12 @@ export class SearchElement extends LitElement {
   }
 
   loadConfig(config) {
-    this.config = config;
+    // Define element attributes outside the ``contructor`` because it's not
+    // allowed to add an element to the DOM with attributes.
+    // See https://stackoverflow.com/questions/43836886/failed-to-construct-customelement-error-when-javascript-file-is-placed-in-head
```
agjohnson, line 78: Ah okay, this makes more sense now that I think about it.

The `render` method, or one of the lifecycle methods like `updated`, is what should alter the host element, the constructor actually shouldn't make any changes besides *setting default values*.

This covers the lifecycle of the element more, and mentions that the constructor does not have access to the DOM attributes yet.
https://lit.dev/docs/v1/components/lifecycle/#assorted-use-cases

agjohnson (approved):
This looks great, and solves the error for me. I get the notification examples in the dev server without any errors.

src/search.js
humitos: should we use `firstUpdate` to set these attributes?

> Use [firstUpdated](https://lit.dev/docs/v1/components/lifecycle/#firstupdated) for initializing private fields from DOM attributes (as the constructor doesn't have access to them). Note that [render](https://lit.dev/docs/v1/components/lifecycle/#render) has already run at this point and your changes might trigger another update lifecycle. If it's imperative that you get access to attribute values before the first render happens, consider using [connectedCallback](https://lit.dev/docs/v1/components/lifecycle/#lifecyclecallbacks), but you'll need to do the extra logic for figuring out the "first" update yourself as [connectedCallback](https://lit.dev/docs/v1/components/lifecycle/#lifecyclecallbacks) can be called multiple times.

agjohnson: That's a good question, I feel like your solution isn't wrong even. I don't know if `firstUpdated` is the most standard place for something like this or not, but if doing the update alongside the other properties in `loadConfig` seems to work, then that seems good for now too. It might help to experiment here to, in case we're just missing something regarding the Lit lifecycle and all the extra events we can hook into. That might just take looking at some level of debugging from Lit and see if it is firing off rendering too many times.

However, I also think this can be merged as it does seem to resolve the error and I don't see anything wrong with doing the updates to properties in one place.

humitos: You posted a link from the v1 version of the docs. However, v2 doesn't have that section anymore 🤷🏼 . I found that `firstUpdated` says:

> Implement firstUpdated() to perform one-time work after the component's DOM has been created.

Which I think it makes sense for our use case: add CSS classes after the element is created in the DOM 👍🏼

humitos: I went ahead and implemented this using `firstUpdated` because it makes more sense to me as it uses the framework internals instead of our own `loadConfig` that was created for a different purpose.

agjohnson: Cool, that sounds good to me too. I still don't know a lot about the internal parts of Lit and the lifecycle events, but it seems when we run into issues with things the solutions have so far been to use lifecycle events more. So, I think you're probably right using the Lit internal bits more I'm sure.

PR #149, opened 2023-09-21
Follow pattern to handle exceptions (`objectPath`)

### LINKED ISSUE #122: Pattern to handle exceptions
Anthony made some suggestions about handling exceptions with `try`/`catch` in

* https://github.com/readthedocs/addons/pull/118#discussion_r1321828523
* https://github.com/readthedocs/addons/pull/118#discussion_r1321859324
* https://github.com/readthedocs/addons/pull/119#discussion_r1321875255

I'd like to find a standard pattern to follow on all the addons that complies with the following ideas, if possible:

* does not involve adding a try/catch each time we want to access a field
* validate the `config.addons.<name>` only once at initialization time and not enable it if it's not valid
* does not break other addons if it fails (e.g. invalid data on `hotkeys` addon doesn't break `search`)
* it's simple to understand and communicate to other team members 😉 

What is a good way to implement this idea? Is this idea good? Are there already standard patterns that I'm not aware of?

humitos:
Closes #122


humitos:
@agjohnson I started testing the pattern we discuss in #122. This is a work in progress PR, but I would like to get some feedback from you about it before continuing with the rest of the addons. The idea here is:

* check for all the mandatory fields required by the addon inside `.isEnabled()`
* use any of the mandatory fields without checking if they exist from inside the addon
* check for optional fields inside the addon, and decide what to do if they are not present

Let me know what you think.

src/analytics.js
```diff
@@ -72,20 +92,25 @@ export class AnalyticsAddon extends AddonBase {
...
     }
   }
 
   static isEnabled(config) {
-    return config.addons && config.addons.analytics.enabled === true;
+    return (
+      objectPath.get(config, "addons.analytics.enabled", false) === true &&
+      objectPath.get(config, "projects.current.slug", undefined) !==
+        undefined &&
+      objectPath.get(config, "versions.current.slug", undefined) !== undefined
```
zanderle, line 113: This check will pass, if `slug=""`, or even `slug=null`. Is that the desired behaviour?

src/flyout.js
```diff
@@ -115,8 +116,10 @@ export class FlyoutElement extends LitElement {
 
   renderVCS() {
     if (
-      !this.config.addons.flyout.vcs ||
-      !this.config.addons.flyout.vcs.view_url
+      objectPath.get(this.config, "addons.flyout.vcs.view_url", undefined) ===
+        undefined ||
+      objectPath.get(this.config, "addons.flyout.vcs.name", undefined) ===
+        undefined
```
zanderle, line 122: Similarly here, this will return false even if `view_url` (or `name`) are `""` or `null`. Is that ok?

zanderle (commented):
I hope you don't mind me leaving 2 clarifying questions :)

src/flyout.js
```diff
@@ -219,7 +210,7 @@ export class FlyoutElement extends LitElement {
 
   render() {
     // The element doesn't yet have our config, don't render it.
-    if (!FlyoutAddon.isEnabled(this.config)) {
+    if (!this.config) {
```
agjohnson, line 213: So, this check will need to be stronger for now, as we are still not accessing all `this.config` attributes in a safe way. The test suite is failing due to relaxing this, as the `render()` method is still accessing attributes in a non-safe way.

We do need a better pattern here, and I think your option for only setting the `this.config` in `loadConfig` when the configuration schema is correct is probably one part of this. We might still need to be defensive around accessing missing data though. A larger part of this is probably adding more defense to attribute access in `render()`, instead of calling the configuration properties directly in the `html` template. That is, use the same pattern for accessing nested attributes instead of:

```
html`<div>${ this.config.project.current }</div>`
```

I actually think we'd want to retain logic similar to checking `FlyoutAddon.isEnabled` either way, as we probably want to share the same check on both the Element and the Addon classes. But, perhaps the changes to `loadConfig` would supersede this too.

Until we have more defense around configuration property access for all attributes, I'd leave the check on `FlyoutAdd.isEnabled()`. We do have a few different ways we could proceed here though.

src/analytics.js
humitos: I think you are right and it makes sense to use `""` as default and check that `""` is not the value, like this:

```javascript
objectPath.get(config, "versions.current.slug", "") !== ""
```

That would be a better protection, right?

src/flyout.js
humitos: > The test suite is failing due to relaxing this, as the `render()` method is still accessing mandatory attributes like `this.config.version.current` directly from the `html` template

I understood that the `render()` would never be called if the `isEnabled()` returns `false` but it seems that I'm missing something on the workflow 🙃

humitos: I think I need to keep thinking on this because I'm a little confused. I'm fine keeping `.isEnabled()` here for now. I don't like that both classes depend each other from this pattern and that we are performing the check twice.

In any case, it seems that you also don't want to have the `.isEnabled()` here, but you don't want to remove it until we find a better pattern. So, we are aligned 👍🏼

humitos: > We do need a better pattern here, and I think your option for only setting the `this.config` in `loadConfig` when the configuration schema is correct is probably one part of this

Yeah, this is my idea.

* Initialize `this.config = null`
* call `if (Addon.isEnabled()): loadConfig(config)`
* trust that `this.config` has all the mandatory fields because it was already validated
* check on optional fields before using them

Does this make sense to you?

src/analytics.js
zanderle: That would be better, but in this case we're not protecting against `undefined` and `null`. Which might be ok, depending on what assumptions we make about `config`

humitos: My assumption when I wrote this was: "if the field exists, its value is valid".

Otherwise, it would be a bug in the backend and I thought we couldn't protect the frontend for all the potential bugs in the backend.

zanderle: Yup, I think that's a sensible assumption. However the backend might send `None`/`null` and this would treat it as a valid slug.

Couldn't we just check if the value here is false-y or not?

src/flyout.js
agjohnson: Well, this doesn't seem *too* far off from where we are currently I guess. Though this is defensive around optional configuration attributes, mandatory attributes are still unprotected. As long as we're still accessing the configuration attributes directly in the templates, mandatory or optional, we'll be left open to throwing exceptions for missing attributes.

Even if we have a mandatory/optional split between configuration attributes, parts of the API structure could change or just be missing for some project edge case, etc.

So what you're suggesting does help with the optional configuration attributes, but I would probably say we still need to protect access for all of the mandatory configuration attributes somehow as well.

Maybe this is all a point for abstracting `this.config` with a layer/function/method for config lookup and always using that instead. I see this being important if we need a lot of logic in template code as we probably want this sort of logic to be reusable by theme authors.

agjohnson: I hit this at https://github.com/readthedocs/readthedocs.org/pull/10831, because the configuration access and html template aren't defensive around access to the attribute or generated build URL.

src/analytics.js
humitos: I don't have a strong opinion here, but I think if we want to be sure the value is valid we should use a JSON scheme to compare against and rely on that to decide whether or not enable the addon.

humitos:
I'm marking this as ready for review. We haven't arrived to a perfect solution yet, but at least we are using `objectPath` in more addons which protects us a little more from accessing to unexpected values from `config`. I'd like to move forward with this pattern for now and continue thinking on different approaches/updates we could apply here.

humitos:
The more I think about this regarding being more strict about data validation, the more I think we would end up doing a JSON schema validation. I think this will be easier to handle and find issues.

Doing a small research I found these two packages that we could use:

* https://www.npmjs.com/package/jsonschema
* https://www.npmjs.com/package/ajv

I think we can probably create each `addons.*` scheme inside the addon itself and validate that part of the JSON there. Besides, we can validate the global part of it like `projects.*` and other global attribute at initialization.

humitos:
There other alternatives like https://nitipit.github.io/edictor/ that we can explore also.

humitos:
> JSON schema validation

This is what we are using on Read the Docs for our `readthedocs.yaml` file: https://github.com/readthedocs/readthedocs.org/blob/9d55392daa1311accc5f86408b9e146ff04ee629/readthedocs/rtd_tests/fixtures/spec/v2/schema.json#L2

agjohnson:
> The more I think about this regarding being more strict about data validation, the more I think we would end up doing a JSON schema validation

Agreed, I think I'm coming to the same conclusion here. It was noble to try to avoid this additional step, but being explicit with the data structure here is clearer, easier, future proof, and probably the most benefit we can provide ourselves.

I don't have experience with ajv, but is seems the most widely used. Either library is probably fine here though. I'm guessing this file would probably be useful to both the addons library and the API response test suite on readthedocs.org (at very least).

humitos:
I was able to codify a small JSON schema and use `ajv` to validate it.

```javascript
ajv.getSchema("addons")({analytics: {code: 1, enabled: true}});
false
ajv.getSchema("addons")({analytics: {code: "UA-1", enabled: true}});
true 
```

I think we can start defining the most important part of the schema and leaves some other parts just as `type: "object"` without specifying all the attributes for the initial approach. After we have it working, we can expand it with more and more keys to make it more accurate.

I will open a PR with this initial approach soon that will close this one since we may not use `objectPath` as we thought.

PR #151, opened 2023-09-21
Packages: call `npm update`

humitos:
Just keep the packages updated.


agjohnson:
Just a heads up, `npm update` might not quite update packages like you might want. Namely, it doesn't update the version listed in package.json.

Folks commonly use `node-check-updates` to keep the package.json updated (which is a silly pattern to rely on an outside package to have to do this, but :shrug:)

agjohnson (approved):
This should be harmless, but noted that it might be worth running `ncu` to get the update version selectors in package.json too.

humitos:
Cool! Doing the wrong thing and getting reviews on them is a good way to learn how to do the right things 😅

I'll check the package you mentioned 👍

PR #152, opened 2023-09-22
Packages: upgrade all with `ncu -u`

humitos:
It modified some .js and .css files because it upgraded the `prettier` version as well, and it seems that trailing commas are now added.

Related #151


package.json
```diff
@@ -56,5 +56,8 @@
         ]
       }
     }
+  },
+  "dependencies": {
```
agjohnson, line 60: I'd install ncu globally (`npm install -g npm-check-updates`), it probably shouldn't be a dependency, or devDependency, of the package

agjohnson (approved):

package.json
humitos: Makes sense. I've removed it.

PR #162, opened 2023-10-10
GHA: run npm run build before creating the PR

humitos:


agjohnson (approved):
Seems good!

Generally speaking, I will say that I'm not sold on the benefits of automating package updates. These PRs (and dependabot PRs similarly) will commonly conflict and without a thorough test suite could be prone to introducing untested error cases.

I'd be extra careful in testing all of these PRs.

PR #173, opened 2023-10-24
DocDiff: use `objectPath` to avoid breaking on `addons.doc_diff`

humitos:
Since we are not always sending the `url=` parameter, it could happen that `addons.doc_diff` does not come in the response. This commit uses `objectPath` to be prepared for that case.

Related #170


humitos:
I need to merge this for today's deploy, but I'm happy to get a review here and adapt with the feedback received. It's pretty simple, tho.

agjohnson (commented):
This change looks fine! :+1:

PR #174, opened 2023-10-25
Notifications: don't show when `stable` is not available

humitos:
https://doc.pypy.org/en/latest/ has only one version named `latest` and we were showing the "You are reading a development version" which is technically true, but doesn't make too much sense.

In those cases, when `stable` version is not available, we do not show a notification at all.

Solves https://github.com/readthedocs/readthedocs.org/issues/10791


jertel (commented):
In the banner message text, there is a spelling typo:
  `publised` should be `published`

And a grammar typo:
  `You may be reading and old version` should be `You may be reading an old version`

I realize these aren't part of this PR's diff but I figured since you already have a PR open you could make the fix here.

humitos:
Thanks, I've opened https://github.com/readthedocs/addons/pull/178 to fix this.

src/notification.js
```diff
@@ -33,8 +33,9 @@ export class NotificationElement extends LitElement {
       external: null,
       stable: null,
     };
-    this.reading_latest_version = null;
-    this.stable_version_available = null;
+    this.reading_latest_version = false;
+    this.reading_stable_version = false;
+    this.stable_version_available = false;
```
agjohnson, line 38: This is mixing variable casing patterns, which isn't ideal. So far, you've been using camel case variable names in this library.

It's worth noting that the JS community is not as standardized this, though camel case is widely used. Snake case would have been a valid choice too though, and the new dashboard was using this casing pattern -- mostly to keep Python and JS familiar though.

If we want to settle on snake casing, I'm +1 on snake case, but -1 on mixing patterns.

Also of note, there is a lot of API data access handled with this library, which is going to be entirely snake cased. This can be accessed using object key selector syntax (`this.config["some_thing"]`) or another pattern, mostly to differentiate the two patterns.

agjohnson (approved):

src/notification.js
humitos: Good point. I've been using camelCase, but it seems that I was writing Python code immediately before jumping into this code 😄 

> Also of note, there is a lot of API data access handled with this library, which is going to be entirely snake cased.

The API response is using snake case with underscores `_`, which allows us to use `this.config.some_thing` --without the need of using `[""]`. Accessing this values use snake case, but when saving them into a variable or creating a new variable, it should be using camelCase on JS code.

I've been following that pattern, but it seems that I broke it here for some reason. I will update this code.

agjohnson: > which allows us to use this.config.some_thing --without the need of using [""]

To clarify, it's valid syntax to access it this way, but accessing API data response properties using string lookup is just a way of separating the two types of variable casing. _Variable casing_ stays camel case but API data properties are accessed using string lookup instead, as the casing system doesn't match.

JS hands out foot guns here by conflating property and attribute access on objects.

humitos: Good point

PR #179, opened 2023-11-02
Make flyout header sticky for when there are a lot of versions

### LINKED ISSUE #128: Flyout: Make the header sticky, when the content is scrollable
As a continuation of https://github.com/readthedocs/addons/issues/121, when there are a lot of versions, the container becomes scrollable. In this situation the header should be sticky, so that it's always visible.

zanderle:
Closes #128


humitos (approved):
Looks good to me!

Peek 2023-11-06 13-07

PR #182, opened 2023-11-08
JSON schema and validation

humitos:
Define a JSON schema and use `ajv` to load/compile it. Then, use this schema to validate the response returned by the API. If the response does not validate the schema we throw an exception and don't load any of the addons.

I plan to modify this to validate each part of the response by the addon itself and disable a particular addon if the validation fails. This will allow having a broken addon without interfering with the rest.

Related #149


src/index.js
```diff
@@ -38,6 +39,13 @@ export function setup() {
         return getReadTheDocsConfig(sendUrlParam);
       })
       .then((config) => {
+        const validate = ajv.getSchema(
+          "https://readthedocs.org/schemas/response.json",
+        );
+        if (!validate(config)) {
+          throw new Error("JSON response does not validate.");
```
zanderle, line 46: Should we just kill all the addons if it doesn't validate? I'm thinking if we ever change the schema: there might be some discrepancies - should we maybe let the addons resolve as much as possible anyway? Or maybe validate per add-on?

zanderle (commented):
I love this approach. I've used it in the past and it's a great way to keep backend and frontend in sync. It's also a great way to write some tests to keep checking if the API response conforms to the schema or not.

src/index.js
humitos: > Or maybe validate per add-on?

Yes. This is my end goal. I want to be able to enable addons if the data required for that addon is validated, even if there are parts of the data returned by the API that's corrupted.

In fact, I don't think we will end up validating that data returned by the API "as a whole" since we may don't care about _all the attributes of it_.

humitos: Following this approach, what I have in mind for analytics addons, for example, it will validate:

- required `projects.current.slug` (string)
- required `versions.current.slug` (string)
- optional `projects.current.language.code` (string)
- optional `projects.current.programming_language.code` (string)
- optional `readthedocs.analytics.code` (string)`

To be able to enable the analytics addon, we don't care if `builds.current.slug` is invalid, for example, or any other field outside of the ones mentioned here 👍🏼

humitos: I added two commits with this approach:

* Analytics: cb9dcad
* EthicalAds: 3c967a7

I will continue with this idea because I like it the most so far.

humitos:
I updated this PR and I'm happier with the approach I'm proposing here. I will continue working on this PR to apply this approach to all the addons. I will also take a look at the testing suite to check how the approach works together with that.

src/analytics.js
```diff
@@ -86,6 +87,9 @@ export class AnalyticsAddon extends AddonBase {
   }
 
   static isEnabled(config) {
-    return config.addons && config.addons.analytics.enabled === true;
+    const validate = ajv.getSchema(
+      "https://readthedocs.org/schemas/addons.analytics.json",
```
agjohnson, line 91: I'm assuming this is a completely arbitrary string, and the URL probably doesn't matter, but I might suggest using a more obviously fake URL in case this somehow translates to a web request at some point -- or perhaps just to help communicate the URL doesn't matter at all.

No preference on what exactly that is, but I'm describing something like:

```suggestion
      "https://v1.schemas.readthedocs.org/addons.analytics.json",
```

src/data-validation.js
agjohnson, line 1: I think this approach could get a bit redundant, there will be a lot of overlap between the schemas -- ie defining `config.projects` and `config.builds` more than once.

I suspect we can get around this by sharing common schemas though.

agjohnson (commented):
This approach is looking great so far and per-addon validation seemingly makes the most sense so far.

src/analytics.js
humitos: He, yeah. This URLs are dumb, but it's how the identifier works 🤷🏼 

> You don't have to host your schema files at the URIs that you use as schema $id. These URIs are only used to identify the schemas, and according to JSON Schema specification validators should not expect to be able to download the schemas from these URIs.

_(from https://ajv.js.org/guide/combining-schemas.html)_

The URI doesn't matter at all. I'm fine using whatever fake URL here 👍🏼

src/data-validation.js
humitos: Yes. I started using `$ref` to share these chunks but I went back because the JSON Schema is pretty verbose already and using `$ref` adds more complexity that we don't need at this point.

Also, even if the `config.projects.current` and `config.builds.current` will be the exact same object no matter what addons we are validating; the addon may require different fields from them. For example, notification addons requires `config.projects.current.single_version` but `analytics` doesn't. In those cases, we won't be able to share these chunks anyways.

Otherwise, not sending `single_version` for whatever reason will make all the addons to be disabled, loosing the idea of "validate per addon".

I'm fine sticking with the current pattern that may duplicate some chunks for now in favor of being explicit, self-contained and useful for per-addon validation. We can come back to this in the future and think about refactoring the validation process if we consider.

humitos:
I'm getting this error when running tests and I'm not able to trace it down. @agjohnson I'd appreciate if you have a clue here that can help me to debug this better.

```
      TypeError: (0 , applicator_1.default) is not a function
        at node_modules/ajv/dist/vocabularies/draft7.js:19:30
```

humitos:
I understand this issue is because of UMD and ES modules, and I understand we are doing some type of module's type conversion when running the tests via rollup, but I'm not sure to understand all of this 🙃

humitos:
It seems the key issue because of the Rollup plugin is doing something that may not be 100% accurate. Commenting this lines https://github.com/readthedocs/readthedocs-client/blob/14a3b874d7f067b8fd69bc451ff0d957459cdb94/web-test-runner.config.mjs#L24-L27 I get a different error:

```
SyntaxError: The requested module './../node_modules/ajv/dist/ajv.js' does not provide an export named 'default'
```

humitos:
I finished this PR regarding the JSON schema and validation. I'm happy with the current pattern. It's strict enough to help us find issues with invalid data and keeping the other addons sane when the data is invalid only for one addon.

I'd like to get some help on making the test suite to run properly, without adding more tests for now, but adapting the existent ones to follow the current approach. I will add tests in a following PR after merging this one.

src/data-validation.js
agjohnson: Yeah agreed. It's refactoring we can always do later, and we can be mindful of this for now at least

agjohnson:
So, the reason why these type of errors come up is generally because of cyclical exports in the library. For the most part, you can get around unclear exports by using:

```
import { default as Ajv } from "ajv";
```

However, in this case, this doesn't quite solve the issue as the exports from the built bundle are a little off already. This is from built bundle:

```
...
module.exports = exports = Ajv;
Object.defineProperty(exports, "__esModule", { value: true });
exports.default = Ajv;
...
```

This is the start of the cyclical export that Rollup and the commonjs plugin seem to have trouble with.

This probably doesn't help explain the solution too much yet, and is mostly a review of what I found setting up WTR. I'll experiment a bit more first, but we should be able to resolve this on our side.

agjohnson:
I took another swing at this. See #185 for the fix to WTR config to make our configuration play a bit nicer with both ajv and visual-dom-diff.

Ultimately, I found the issue because most of the `require` calls in ajv were expecting the return to include the whole `module.exports` -- that is your error above, the expected return from the `require()` needed to be:

```js
applicator_1 = {default: getApplicator}
```

Instead, with Rollup's commonjs plugin, and the `preferred` mode to [requireReturnsDefault](https://www.npmjs.com/package/@rollup/plugin-commonjs#requirereturnsdefault), the return was:

```js
applicator_1 = getApplicator
```

Specifically, the module exports were squashed because the intention would never to use `applicator_1.default` directly.

I got new test failures, but they look like exactly what I would expect. The new failures are because we are only passing in partial configuration objects and with your change we can no longer do that.

humitos:
This PR is ready to review and merge. I'm happy with the pattern and I'm feeling more comfortable with it. There are things we will want to refactor in the near future, but as I said before, I wanted to be pretty explicit at this point since it's pretty easy to make mistakes with the JSON schema.

In the meanwhile, I will open another PR based on this one to write some extra test cases 👍🏼

src/data-validation.js
agjohnson, line 1: Just a note here, I didn't inspect the actual schemas here deeply. I think we can adjust if they don't match what we are outputting from the application and don't need to be strict about ensuring they are strictly accurate yet.

agjohnson (approved):
This all looks great and it all feels easier to work with now :+1:

src/data-validation.js
zanderle: Are we planning to add some backend tests to see if the API response conforms to the schema(s)?

humitos: @agjohnson 

> Just a note here, I didn't inspect the actual schemas here deeply. I think we can adjust if they don't match what we are outputting from the application and don't need to be strict about ensuring they are strictly accurate yet.

Yes. Currently, they are "good enough" but not perfect. I've been finding issues while writing the test cases, so I expect to find more as we keep working on the addons. We already have the pattern defined, so updating them shouldn't be too hard

@zanderle 

> Are we planning to add some backend tests to see if the API response conforms to the schema(s)?

This is something that I'd eventually want but we can't prioritize right now. We will need to find a way to share the JSON schema between the two repositories and execute the tests in both platforms with the same schema. "Sharing the schema" probably means writing the schema in JSON instead of JS that's something I tried to avoid because JSON doesn't support comments among others painful things, like not supporting trailing comma 🙃 

Do you have experience doing this?

zanderle: The way I did it in the past was to have the schema in json and then use `description` and [`$comment`](https://json-schema.org/understanding-json-schema/reference/comments) keywords to provide context where needed.

It's been a few years since I did this, so I can't remember how exactly I shared it between the frontend and backend repos. Let me try to find that information actually.

Issue #147, opened 2023-09-21
Search: improve the look&feel of the modal


humitos:
Currently, the modal looks "fine". However, it's pretty basic and it could be improved in a lot of ways. I'm opening this issue here to get some feedback about _what are those improvements we can do?_ related to look&feel, style, content, etc.

humitos:
There is some extra context in https://readthedocs.slack.com/archives/C029M3W01KR/p1696954421985799?thread_ts=1696940708.523109&cid=C029M3W01KR

PR #184, opened 2023-11-13
Development: mock/hardcode a search response

humitos:
Mocked/Hardcoded response from Search V3 API to be able to perform a search and get some results as examples. Useful for #147


src/search.js
```diff
@@ -407,9 +407,14 @@ export class SearchElement extends LitElement {
...
 
     let deboucedFetchResults = () => {
-      const url =
+      let url =
         API_ENDPOINT + "?" + new URLSearchParams({ q: query }).toString();
 
+      // Retrieve a static JSON file when working in development mode
+      if (window.location.href.startsWith("http://localhost")) {
+        url = "/_/readthedocs-search.json";
+      }
```
agjohnson, line 416: At some point we will need a better pattern for these overrides. This conditional won't always be correct or could be triggered outside addons development. We can control conditional variables/globals in the Webpack configuration, using a few different patterns.

agjohnson (approved):

src/search.js
humitos: Yes. I tried using IS_PRODUCTION webpack variable but doesn't work anymore for some reason 🤷‍♂️

humitos: I opened https://github.com/readthedocs/addons/issues/186 so we don't forget about this.

PR #185, opened 2023-11-14
Alter the configuration around externals and default exports

agjohnson:
Preferred default export mode is not required anymore, and this plays
nicely with other TS libraries.

Refs #182


humitos (approved):
This is great! Thanks for checking this and solving the problem. I've tested in my branch with ajv and it works great!

PR #187, opened 2023-11-14
Test: HTML notification test cases

humitos:
- Test the title of the "latest" notification
- Test the title of the "non stable" notification
- Test the title of the PR notification
- Test default CSS classes

The main point of this PR is to define the pattern we will use to check logic inside the browser and test HTML attributes, etc.

Closes #149


tests/notification.test.html
```diff
@@ -0,0 +1,115 @@
...
+          it("on stable version", async () => {
+            // Set current version as non-stable and non-latest
+            config.versions.current.slug = "v3.x";
+            config.versions.current.type = "tag";
+
+            const addon = new notification.NotificationAddon(config);
+            const element = document.querySelector("readthedocs-notification");
+
+            // We need to wait for the element to renders/updates before querying it
+            await elementUpdated(element);
```
agjohnson, line 62: Just noting because this is commented, but almost every test case will have some async await to it. This is all expected and normal.

The comment is completely accurate, but made it sound like `await` here is a hack. You can expect to wait on element render or you can use promises directly too, if you aren't finding that isn't enough.

tests/notification.test.html
```diff
@@ -0,0 +1,115 @@
...
+      import { expect, elementUpdated } from "@open-wc/testing";
+      import { runTests } from "@web/test-runner-mocha";
+      import * as notification from "../src/notification.js";
+
+      // Define the config globally to be able to override on ``beforeEach`` and inside each test
+      let config;
+
+      runTests(async () => {
+        beforeEach(() => {
+          config = {
```
agjohnson, line 13: This method for handling the configuration object is great, this is a good pattern to reproduce on other tests.

agjohnson (approved):

tests/notification.test.html
humitos: Oh, I wrote the comment as documentation to myself because I hit the `element.shadowRoot === null` issue and took me some time to realize that the element was not rendered at the time I was querying it. I understand that `elementUpdated` is the normal flow here and it's OK to me as well. It's definitely not a hack 😄 

> or you can use promises directly too

Promises are things that I don't make me feel super comfortable yet 😅 --so, the current pattern is clearer to me 😄

tests/notification.test.html
humitos: Wohoo! 💪🏼

PR #188, opened 2023-11-15
EthicalAds: support explicit placement

humitos:
Use the `#ethical-ad-placement` explicit placement on Read the Docs to keep backward compatibility with current integrations.

Users just need to define a `div` whenever they want in the HTML and style with the CSS classes they want as follows:

```
<div id="ethical-ad-placement" class="ad-flat"></div>
```

This will tell our EthicalAdsAddon to load the ad in that position. Note that `data-ea-manual="true"` should not be used on these cases.

Closes https://github.com/readthedocs/meta/issues/132


davidfischer (approved):
This looks right but my main concern is around timing and order. Specifically, the ad client code is usually loaded async which is why RTD used to set `data-ea-manual=true` to ensure that the ad client didn't run until after keywords/publisher/excluding-paid was set with RTD's data. I've been following the addons changes but I don't think I fully understand that timing and order there but if these will be handled correctly then this should be fine.

humitos:
> This looks right but my main concern is around timing and order. Specifically, the ad client code is usually loaded async which is why RTD used to set `data-ea-manual=true` to ensure that the ad client didn't run until after keywords/publisher/excluding-paid was set with RTD's data.

In the addons code we are not using `async=true` when loading the js script (we probably should, tho). See https://github.com/readthedocs/readthedocs-client/blob/58ca70115a1d188dd4a87386c5a134d4ee1b00d6/src/ethicalads.js#L100-L108

In any case, the ad placement is injected in sync (no pomises or similar) _before_ loading the js script. So, it's not possible to have an ad placement without the publisher defined.

That said, I still don't understand why `data-ea-manual=true` is used in RTD when there is no explicit call to `ethicalads.load_placement()` anywhere in the source code 🤔 . Can you expand on that?

humitos:
Another question: should we keep using `data-ea-manual=true` for the ad placement explicitly defined by users? should we use `data-ea-manual=true` for the ad placement injected automatically when it's not already present on the HTML?

humitos:
OK! That makes a lot more sense. I was searching for `load_placement` that's why I didn't find it 😅 

I updated the code to follow what we have in that Read the Docs .js file:

* inject the ethicalads script with `async=true`
* subscribe to `load` event on that script tag injected
* call `ethicalds.load()`

Also, I'm always setting `data-ea-manual=true` even if the user has defined the `<div id="ethical-ad-placement"></div>`.

humitos:
I disabled the auto-inject ad to only support explicit ad placement defined by users for now, as we talked. We can uncomment the code once we want to move forward with that.

With this latest changes, this PR is ready to be merged and deployed next week.

PR #192, opened 2023-11-17
Test: analytics with mocked requests

### LINKED ISSUE #150: Tests: write basic tests for each of the addons
We have pretty basic tests using `web-test-runner` for all the addons. The initial structured is set, which is not minor 😄 . There work we want to do here is:

- [x] define the patter to initialize that addons with the correct config
- [x] write small test to, based on their config, different things happens
- [x] check some HTML content based on the config used to initialize the addon

> Hint: Anthony is doing some good progress on the ext-theme repository that we can copy from here.

humitos:
After researching a lot I found the pattern that I like the most.

* it uses `sinon`, which is a standard library/framework for mocking
* it uses `sinon.fakeServer` to mock requests/responses easily, giving us a lot of extra features to inspect the mocked URL and assert multiple fields
* since `sinon` doesn't support (yet) mocking global `fetch`, I'm using `unfetch`, which is a small library that converts `fetch` into `XMLHttpRequest` call by imitating the same API than global `fetch`
* once `sinon` add support for this, we just need to remove an import

I'm happy with this pattern and I will keep expanding it as we go. Nothing is written in stone here, but I've tested all the other different patterns we talked about and they are are dirty to me, so I wanted to keep it simple here for now while having the `sinon.fakeServer` features that I like.

Closes #150


humitos:
I'm merging this PR since it uses the same approach than in https://github.com/readthedocs/addons/pull/196 and it's already approved there. That was the only noticeable thing from this PR.

I'm happy to make improvements changes on the test code, but I'm pretty happy with the sinon/unfetch mock pattern so I'm sticking to it.

PR #196, opened 2023-11-21
 Test: entry point (get config and enable addons) 

humitos:
Follow the pattern from #192 to write more tests.

* Update data schema validation for notification. It requires some fields to be present on the JSON.
* Remove invalid logging on tests from notification.
* Update addon name on notification.
* Create a constant for `ADDONS_API_ENDPOINT` so it can be shared.
* Create some tests for `readthedocs.setup()` call.


src/readthedocs-config.js
```diff
@@ -31,12 +36,14 @@ export function getReadTheDocsConfig(sendUrlParam) {
...
 
+  // TODO: find a way to disable this when running on test environment
+  //
   // Retrieve a static JSON file when working in development mode
-  if (window.location.href.startsWith("http://localhost")) {
-    url = "/_/readthedocs-addons.json";
-  }
+  // if (window.location.href.startsWith("http://localhost")) {
+  //   url = "/_/readthedocs-addons.json";
+  // }
```
agjohnson, line 46: Yup, I was expecting we'd hit this scenario pretty quickly in testing. I mentioned a few options in the past -- global variable set via Webpack, global replace set via Webpack, or reuse the Webpack mode global variable and only support this URL when using `npm run dev`

tests/index.test.html
```diff
@@ -0,0 +1,144 @@
...
+      import { runTests } from "@web/test-runner-mocha";
+      import {
+        ADDONS_API_VERSION,
+        ADDONS_API_ENDPOINT,
+        CLIENT_VERSION,
+      } from "../src/utils.js";
+      import * as docdiff from "../src/docdiff.js";
+      import * as readthedocs from "../src/index.js";
+
+      /* import { default as chai } from "chai";
```
agjohnson, line 17: `chai` is not surfaced directly, it is surfaced through `@open-wc/testing`

tests/index.test.html
```diff
@@ -0,0 +1,144 @@
...
+            expect(server.requests[0].url).to.be.equal(url);
+          });
+
+          it("skip everything because Read the Docs Embed is present", async () => {
+            const script = document.createElement("script");
+            script.src = "/_/static/javascript/readthedocs-doc-embed.js";
+            document.body.append(script);
+
+            // Our .setup() returns "false" when old embed js is present
+            const result = readthedocs.setup();
```
agjohnson, line 137: Oh, this is a little confusing. `readthedocs.setup()` should probably be always async, and in this case should `return Promise.reject()` instead. Hybrid async/sync methods are prone to weird bugs, which you might be able to notice if you call `await readthedocs.setup()` and it returns `false` instead of a `Promise`.

agjohnson (approved):

src/readthedocs-config.js
humitos: I tried the Webpack approach but it doesn't seem to work as expected. I understand we are not using webpack to run the tests and that's why the `webpack.DefinePlugin` doesn't work there (I tested this by adding another variable named `IS_TESTING` at https://github.com/readthedocs/readthedocs-client/blob/2a6ed8736e580cb2ea7661efac187c6e7e3b9610/webpack.config.js#L76-L78).

However, I'm not 100% sure how everything connects together 🤷🏼

tests/index.test.html
humitos: Oh, yeah, this is part of some testing I was doing where I wanted to add extra capabilities to chai via plugins and `chai.use(...)`

src/readthedocs-config.js
humitos: Do you know a way to differentiate "running tests" from "running development"?

humitos:
Interesting, we have already a good coverage 🎉 

```
Code coverage: 77.14 %
View full coverage report at coverage/lcov-report/index.html
```

src/readthedocs-config.js
humitos: Found it! We use `web-dev-server` and we can use the `rollup-replace` plugin there: https://modern-web.dev/docs/dev-server/writing-plugins/examples/#constants-replacement

I've applied this in a64179c

tests/index.test.html
agjohnson: I put up an example at #202 but it might need a bit more testing, feel free to poke at it there.

PR #198, opened 2023-11-22
API: use `version_schema` instead of `single_version`

### LINKED ISSUE #195: Use the new `project.versioning_schema` API field
We are using `project.single_version` in different places, e.g: https://github.com/readthedocs/readthedocs-client/blob/a9f2221d74626b9db671f9a134613fd050c1727f/src/flyout.js#L70

However, we changed the API in https://github.com/readthedocs/readthedocs.org/pull/10845/. We should update our code to match the new implementation.

humitos:
The backend API response changed and `single_version` is deprecated now.

Closes #195


public/_/readthedocs-addons.json
```diff
@@ -14,7 +14,7 @@
       "programming_language": {
         "code": "words"
       },
-      "single_version": false
+      "version_schema": "multiple_versions_with_translations"
```
stsewd, line 17:
```suggestion
      "versioning_scheme": "multiple_versions_with_translations"
```

src/data-validation.js
```diff
@@ -167,10 +167,15 @@ const addons_flyout = {
       properties: {
         current: {
           type: "object",
-          required: ["slug", "single_version"],
+          required: ["slug", "versioning_schema"],
           properties: {
             slug: { type: "string" },
-            single_version: { type: "boolean" },
+            versioning_schema: {
```
stsewd, line 173: Same for all other places

```suggestion
            versioning_scheme: {
```

humitos:
I updated the field name. Thanks @stsewd for pointing this out 💯

agjohnson (approved):

PR #200, opened 2023-11-22
Development: add Material for MkDocs as example

humitos:
I added Material for MkDocs example to have it hand. I had to fix some small pre-commit issue.


agjohnson (approved):
Looks good, seems like this is mostly boilerplate.

It might be helpful to document how this example was created, so that we understand how to update it in the future and what pieces are critical for addons.

humitos:
> It might be helpful to document how this example was created, so that we understand how to update it in the future and what pieces are critical for addons.

Yes. I wasn't sure what to do here and I went for the quick and simple direction for now: upload a built site 😅 

I would like to have the sources and the resulting HTML built together, but I thought I was kind of duplicating what we have in `test-builds` [^1]. So, I thought that we should probably find a way to integrate both repositories instead of duplicating the work.

I'd like to think a little more about this idea. Maybe it's just a Git submodule with a README with instructions on how to build each of them.

[^1]: by the way, this example was created with the `mkdocs-material` branch from `test-builds` 😄

humitos:
I included these 3 examples here to have them at hand to do quick tests, but I'm happy to expand the idea to pattern that works better.

humitos:
I opened https://github.com/readthedocs/addons/issues/203 to think more about this.

humitos:
> It might be helpful to document how this example was created,

Meh, I forgot that I have also pushed the sources to `examples/` directory. So, I should do the same for MkDocs since they are a few set of files.

PR #202, opened 2023-11-22
Always return a promise from readthedocs.setup()

agjohnson:
Refs #196


tests/index.test.html
```diff
@@ -124,14 +124,16 @@
...
             document.body.append(script);
 
-            // Our .setup() returns "false" when old embed js is present
-            const result = readthedocs.setup();
-            expect(result).to.be.false;
+            return readthedocs
+              .setup()
+              .catch((error) =>
+                expect(error.message).to.equal("Skipping injection"),
+              );
```
agjohnson, line 136: What I did here was remove the `async` test execution and treated the response as a promise object. The return statement avoids the test timeout that would normally occur here.

humitos:
I created a PR following your example at https://github.com/readthedocs/addons/pull/204 and using a Chai plugin for promises 👍🏼

tests/index.test.html
humitos: I read a little about this approach and I found that with this type of tests there is problem when the function (e.g. `readthedocs.setup()`) works fine since the `.catch` will never be executed and the test will report a false positive here.

agjohnson:
Sounds good, either option should be enough for this test.

PR #204, opened 2023-11-23
Always return a promise from readthedocs.setup()

humitos:
I followed Anthony's example from #202 to always return a Promise.

I knew there was a way to use `chai` via a plugin to test Promises. So, I researched a little more about this and I found the exact plugin I've seen before: [`chai-as-promised`](https://www.chaijs.com/plugins/chai-as-promised/)

From the first example, you can see it's used to express exactly what you wrote in a more simple way and providing some extra helpers.

I did a quick test to fail the check using a different error message and it worked great:

```
tests/index.test.html:

 ❌ Main library > skip everything because Read the Docs Embed is present
      AssertionError: expected promise to be rejected with 'Error: Skip' but it was rejected with 'Error: Skipping injection'
      + expected - actual
      
      -Error: Skipping injection
      +Error: Skip
```

I like this pattern since it's pretty explicit and it uses the community standards.


tests/index.test.html
```diff
@@ -124,16 +130,19 @@
             expect(server.requests[0].url).to.be.equal(url);
           });
 
-          it("skip everything because Read the Docs Embed is present", () => {
+          // The last call of "done" function is required here because is how Mocha works with async code.
+          // Read more at https://mochajs.org/#asynchronous-code
+          it("skip everything because Read the Docs Embed is present", function (done) {
```
agjohnson, line 135: Does this not work with arrow functions?

```suggestion
          it("skip everything because Read the Docs Embed is present", (done) => {
```

tests/index.test.html
```diff
@@ -124,16 +130,19 @@
...
             const script = document.createElement("script");
             script.src = "/_/static/javascript/readthedocs-doc-embed.js";
             document.body.append(script);
 
-            return readthedocs
+            // NOTE: there is no need to use "expect" here when using "Promise.should"
+            // Read more at https://www.chaijs.com/plugins/chai-as-promised/
```
agjohnson, line 141: Just noting that expect/should are part of the same API:

https://www.chaijs.com/api/bdd/

`expect` contains most of the assertions, and `should` is more exclusively just testing error catching. Not sure why there is a split here in the API though :shrug:

agjohnson (approved):

PR #206, opened 2023-11-28
Hotkeys: do not trigger them on `textarea` HTML elements

### LINKED ISSUE #205: Docdiff should not apply the 'd' hotkey in textareas
Docdiff doesn't apply the `d` hotkey when typing in `<input>` fields but should apply similar checks to `<textarea>` fields.

https://github.com/readthedocs/addons/blob/ff8baf238d4734152dc2373afac0757a4dc7c6f5/src/hotkeys.js#L49-L69

You can see this bug in the [notes field](https://www.ethicalads.io/publishers/#inbound-form) on ethicalads.io where the 'd' character cannot be typed.

humitos:
Closes #205


davidfischer (approved):
This looks like it should work. I'm sure there's some people out there doing weird things so the sooner we make this really easy to disable the better.

humitos:
> I'm sure there's some people out there doing weird things so the sooner we make this really easy to disable the better.

Yeah, definitely. We are tracking this at https://github.com/readthedocs/ext-theme/issues/211

If you want me to disable the hotkeys in your project, I can do that from the Django admin for now.

Issue #210, opened 2023-12-06
Search still shows up in the flyout when it's disabled


ericholscher:
I tried to disable search for a user, but it still shows up in the flyout. I'd expect disabling search would remove it from the flyout.

PR #214, opened 2023-12-11
Don't show search input on flyout if search is disabled

humitos:
If `addons.search.enabled: false`, the search input that triggers the search modal from the flyout is not added to it.

|Search Enabled|Search Disabled|
|-------|-------|
|Screenshot_2023-12-11_16-14-13|Screenshot_2023-12-11_16-14-28|

Closes https://github.com/readthedocs/addons/issues/210


tests/flyout.test.html
```diff
@@ -215,7 +218,19 @@
...
+
+          it("snapshot flyout with search disabled", async () => {
+            // Disable the search from the config object
+            config.addons.search.enabled = false;
+
+            /* console.log(document.querySelectorAll("readthedocs-flyout").length); */
+            const addon = new flyout.FlyoutAddon(config);
+            const element = document.querySelector("readthedocs-flyout");
+            await elementUpdated(element);
+            await expect(element).shadowDom.to.equalSnapshot();
```
agjohnson, line 233: The snapshot pattern is neat, but I'm seeing now how this sort of test pattern is not very explicit. We have to trust that the test snapshot output is exactly what we want if we rely on just this test. I still see snapshot tests being helpful as secondary tests, informing us that the output didn't change in a PR. But I do see how relying on snapshots for a test like this -- ensure the search input is not in the output -- is not really reflected at all in this test unless you really dig, or know to dig, through the output HTML.

So, I'd probably at least add to this test case an explicit check for the search input, or write a primary test case for this check.

agjohnson (approved):
Changes look good here, I only noted a test case pattern that I think we should improve on as we go.

tests/flyout.test.html
humitos: Yeah, I noticed the same about the test case not being super explicit. However, when it fails, the diff shown is pretty clear and it's easy to realize what's wrong.

Manually comparing a hardcoded chunk of HTML in the JS with the DOM as I did in the other test cases, is probably worse than this.

However, a combination of an snapshot with a simple "the search input is not there" test is probably the best --as you mentioned.

agjohnson: Yeah, I'd agree that snapshots is a better approach than manually testing. Though I can see automatic snapshots drifting over time too, it might be easy to miss updates to these. Some basic, explicit test cases help keep both in check though.

PR #228, opened 2024-01-15
DocDiff: enforce data validation

humitos:
When we worked on the data validation, we broke DocDiff because we weren't taking parameters (`baseUrl` and `rootSelector`) from the API response.

This commit enforces them being present in the API response and used when performing the request/fetch of the old version of the current page.


src/docdiff.js
```diff
@@ -2,7 +2,7 @@ import { ajv } from "./data-validation";
 import styleSheet from "./docdiff.css";
 import docdiffGeneralStyleSheet from "./docdiff.document.css";
 
-import * as visualDomDiff from "visual-dom-diff";
+import { default as visualDomDiff } from "visual-dom-diff";
```
agjohnson, line 5: I had struggled with this particular change trying to get web-test-runner working correctly. We may find that the `import *` syntax is still required at some point, though tests seem to have passed okay.

agjohnson (approved):

src/docdiff.js
humitos: He, I had problems running tests with `import *` and I changed to the this because I worked 🤷🏼

PR #232, opened 2024-01-16
Data validation: allow `multiple_versions_without_translations`

humitos:
This is a valid value that I missed when I upgraded this code.

Solves issue reported at https://github.com/pypa/pipx/issues/698#issuecomment-1894008673


agjohnson (approved):

humitos:
I added a small unit test for this case and I'm merging it now.

PR #234, opened 2024-01-22
Docdiff: change how `visualDomDiff` is imported

humitos:
Using `import { visualDomDiff } from "visual-dom-diff";` makes the code to work properly, but tests fail.

However, using `import { default as visualDomDiff } from "visual-dom-diff";` the code fails but the test pass.

What is the way to make both work? cc @agjohnson 

Peek 2024-01-22 17-42


src/docdiff.js
```diff
@@ -2,7 +2,7 @@ import { ajv } from "./data-validation";
 import styleSheet from "./docdiff.css";
 import docdiffGeneralStyleSheet from "./docdiff.document.css";
 
-import { default as visualDomDiff } from "visual-dom-diff";
+import { visualDomDiff } from "visual-dom-diff";
```
agjohnson, line 5: I noted this on the last PR that you changed, I think `import * as visualDomDiff` was very intentional. I had tried all of the permutations to get tests running. Is there a reason that did not work?

agjohnson: I would start at that error and work backwards, and inspect what `visualDomDiff` actually is right before the exception is thrown. I know from working with this library that the module imported is effectively something close to this:

```js
class visualDomDiff { ... }

visualDomDiff.visualDomDiff = visualDomDiff;

export visualDomDiff;
```

This tries to be backward compatible so that these both work:

```js
const visualDomDiff = require('visual-dom-diff');
import { visualDomDiff } from "visual-dom-diff";
```

Unfortunately, this is also a cyclical reference, and causes headaches in different ways for Rollup/Webpack. This is a simplification and there is more that happens here, but I can't quite explain the internal parts.

What is ultimately happening here is that visual-dom-diff is a TypeScript library, with it's own compilation step to JS. With libraries like this, we have to rely on the compile time output type the library was configured with, which can be various flavors of ES modules or commonjs modules.

Between this and hacks used to support both commonjs and ES module imports, things get really _weird_.

humitos:
I was able to import `visualDomDiff` in a way that works for testings and also works locally. I'll release a new patch version of the addons and test it on production for since we've got a report in #263 that it's not working.

humitos:
The library import works 💯 . See example in https://github.com/readthedocs/addons/issues/263#issuecomment-1994225048

PR #242, opened 2024-02-08
Flyout: keeps the page when switching version/language

### LINKED ISSUE #130: Flyout: keep version on lang switch
## Details

* Read the Docs project URL: https://glpi-install.readthedocs.io/en/develop/install/advanced-configuration.html

Currently, the above page exists only in the develop version of the documentation. 
When switching to another language; current version should be kept but is not (just try to switch to another language).

## Expected Result

User is redirected to the current page on **current** version with selected lang.

## Actual Result

User is redirected to the current page on **latest** version with selected lang (page that does not exists).

### LINKED ISSUE #131: Flyout: keep the anchor when changing the language
Currently anchor in the URL is always removed while changing the language through the flyout menu.

### LINKED ISSUE #211: Flyout: link to change version drops page name
Noticed here:

https://docs.astropy.org/en/stable/coordinates/inplace.html

The versions should all link to `/en/{version}/coordinates/inplace.html`. They link to `/en/{version}/` instead.

- [ ] Page name is dropped from version switch links
- [ ] Page name is dropped from locale switch links (in theory?)
- [ ] Latest version is used instead of current version for locale links

humitos:
Closes #211
Closes #131 
Closes #130


src/flyout.js
```diff
@@ -187,6 +187,30 @@ export class FlyoutElement extends LitElement {
...
+    //   URL: https://docs.readthedocs.io/en/latest/
+    //   Filename: /index.html
+    //   Flyuout URL: https://docs.readthedocs.io/en/latest/
+    //
+    //   URL: https://docs.readthedocs.io/en/stable/
+    //   Filename: /guides/access/index.html
+    //   Flyuout URL: https://docs.readthedocs.io/en/stable/guides/access/
+    return (
+      url.replace(/\/+$/, "") +
+      metaFilename.content.replace(/\/index.html$/, "")
```
stsewd, line 210: I think you need to replace /index.html with /, since the slash is included in the replace.

```suggestion
      metaFilename.content.replace(/\/index.html$/, "/")
```

src/flyout.js
```diff
@@ -187,6 +187,30 @@ export class FlyoutElement extends LitElement {
     `;
   }
 
+  _getFlyoutLinkWithFilename = (url) => {
+    // Get the resolver's filename returned by the application (as HTTP header)
+    // and injected by Cloudflare Worker as a meta HTML tag
+    const metaFilename = document.querySelector(
+      "meta[name='readthedocs-resolver-filename']",
+    );
```
agjohnson, line 195: Is it worth returning something by default if the meta tag is missing for some reason?

src/flyout.js
```diff
@@ -187,6 +187,30 @@ export class FlyoutElement extends LitElement {
...
+    // Examples:
+    //
+    //   URL: https://docs.readthedocs.io/en/latest/
+    //   Filename: /index.html
+    //   Flyuout URL: https://docs.readthedocs.io/en/latest/
+    //
+    //   URL: https://docs.readthedocs.io/en/stable/
+    //   Filename: /guides/access/index.html
+    //   Flyuout URL: https://docs.readthedocs.io/en/stable/guides/access/
+    return (
```
agjohnson, line 208: The standard way to work with URLs in Node is to use `URL()`:

https://nodejs.org/api/url.html#new-urlinput-base

It is similar in effect, but handles much more of the weird, URL specific edge cases that concatenation might miss.

agjohnson (approved):

src/flyout.js
humitos: Hrm, good point. In case the meta tag is not found, we should just generate the links without the filename.

agjohnson: This also just might not be important. I assume as long as the Cloudflare injection is working properly, we don't have to worry about this missing tag.

src/flyout.js
humitos: Good point!

src/flyout.js
humitos: I updated the code to use `new URL()` with a base and filename.

humitos: Hrm, it seems this needs more work.

```
base = "/es/"
"/es/"
filename = ""
""
new URL(filename, base)
Uncaught TypeError: URL constructor: /es/ is not a valid URL.
```

I think that `URL` needs a full `base` URL with scheme and everything. Besides, the `base` doesn't have the version on it, which means that `/es/${filename}` won't be a valid URL.

humitos:
I'm blocking this PR because we need to fix the construction of `new URL` first.

src/flyout.js
agjohnson: Yes, if `input` is relative, `base` is required -- `base` just needs to be scheme and hostname.

You can get this from the window location if you don't have anything more accurate:

```js
let someUrl = new URL(window.location.href);
```

You might still need to append "/" with a string operation though, as URL doesn't quite do what you're doing here either:

```
>  (new URL("/es/latest", window.location)).href;
< "https://docs.readthedocs.io/es/latest"
> (new URL("./index.html", new URL("/es/latest", window.location))).href;
< "https://docs.readthedocs.io/es/index.html"
```

String operation is mostly okay here then, but I would be aware of `URL()` for other cases where you need to handle URLs.

humitos: OK. I found the test were using invalid data for `addons.flyout.versions` and `addons.flyout.translations` because the URLs include there are absolute in the API response, but we were using relative ones in the tests. This is because it was like that originally.

So, I updated the tests and the mocked data we use under `localhost:8000` as well.

src/flyout.js
```diff
@@ -195,6 +195,37 @@ export class FlyoutElement extends LitElement {
...
+    //   Flyuout URL: https://docs.readthedocs.io/en/stable/guides/access/
+
+    // Keep only one trailing slash
+    const base = url.replace(/\/+$/, "/");
+
+    // 1. remove initial slash to make it relative to the base
+    // 2. remove the trailing "index.html"
+    const filename = metaFilename.content
+      .replace(/^\//, "")
+      .replace(/index.html$/, "");
```
stsewd, line 224: Otherwise this may introduce some bugs for paths that end with index.html `foo/test-index.html`

```suggestion
      .replace(/\/index.html$/, "/");
```

But should probably come before the first replace, or a path like `/index.html` won't work.

humitos: Hrm, good point. I reverted this change because `.replace(/\/index.html$/, "/")` doesn't work as expected in the local tests. I will take another look at this.

humitos: Thanks for the review on this. I fixed it on b5b23b3

PR #244, opened 2024-02-09
POC Recent searches

### LINKED ISSUE #33: Search: initial help when the search modal is opened
This may not be _required_ and we could use different approaches here. However, I just thought it could bring a kind of "welcoming message" to have some small help in the initial dialogue of the search modal:

Screenshot_2023-04-04_00-04-22

"Starting typing to find out results. You can use the following syntax patterns"

I'm thinking we can expose something similar to what we are showing when there are no results at all: https://github.com/readthedocs/readthedocs-client/issues/19#issuecomment-1490175025

Algolia shows "No recent searches" there when there is nothing to show or, the recent searches themselves if there some.

Screenshot_2023-04-04_00-07-03

zanderle:
A potential idea to close #33 

This obviously needs some polish when it comes to styling, and a few implementation details. But what do you think of the overall concept? I tried following the Algolia example when it comes to behaviour (storing only the searches that were actually clicked).


src/search.js
```diff
@@ -73,6 +74,7 @@ export class SearchElement extends LitElement {
     this.triggerKeycode = 191;
     this.triggerSelector = null;
     this.triggerEvent = "focusin";
+    this.recentSearchesKey = "readthedocs_search_recent_searches"
```
humitos, line 77:
```suggestion
    this.recentSearchesLocalStorageKey = "readthedocs-search-recent-searches"
```

Just make the variable's name more explicit and use `-` instead.

src/search.js
```diff
@@ -273,6 +275,7 @@ export class SearchElement extends LitElement {
     return html`
       <a
         @mouseenter=${this.mouseenterResultHit}
+        @click=${() => this.storeRecentSearch(block, result)}
```
humitos, line 278: Why are you using an anonymous function here? Is this because you need to pass the attributes `block` and `result`?

src/search.js
```diff
@@ -284,6 +287,52 @@ export class SearchElement extends LitElement {
...
+
+  getRecentSearches() {
+    const recentSearchesString = localStorage.getItem(this.recentSearchesKey);
+    return recentSearchesString ? JSON.parse(recentSearchesString) : [];
+  }
+
+  storeRecentSearch(block, result) {
+    const recentSearches = this.getRecentSearches();
+    // TODO Avoid duplicates
+    // TODO Maybe add some limit to the total number of recent searches
```
humitos, line 331: :100:

src/search.js
```diff
@@ -284,6 +287,52 @@ export class SearchElement extends LitElement {
...
+    }
+    console.log(recentSearches);
+    const listIcon = icon(faClockRotateLeft, {
+      title: "Result",
+      classes: ["header", "icon"],
+    });
+
+    return html`
+      <div class="hit">
+        <p>Recent Searches:</p>
```
humitos, line 303:
```suggestion
        <p>Recent:</p>
```

src/search.js
```diff
@@ -284,6 +287,52 @@ export class SearchElement extends LitElement {
...
+        ${recentSearches.map(
+          ({block, result}) =>
+            html`<div class="hit-block">
+              <a class="hit-block-heading" href="${result.path}">
+                <i>${listIcon.node[0]}</i>
+                <h2>${result.title} ${this.renderExternalProject(result)}</h2>
```
humitos, line 309: Is there a particular reason why you are calling `renderExternalProject` (note the `External` part)?

src/search.js
```diff
@@ -284,6 +287,52 @@ export class SearchElement extends LitElement {
...
+              </a>
+
+              ${html`${this.renderBlockResult(
+                    block,
+                    1,
```
humitos, line 314: This number needs to be dynamic, I guess.

humitos (commented):
This looks great for an initial implementation 🎉 . Thanks for putting all of this together 👍🏼 

I did a small review of the changes and also tested locally and it worked great!

I'd add to the TODO list as well:

- a way to clean all the recent searches
- a way to remove one item from recent searches

Let's continue in this direction.

src/search.js
zanderle: Exactly. If I just did `this.storeRecentSearch(block, result)` it would get evaluated/called as soon as the component is created. This way it gets called only on click.

src/search.js
zanderle: I just reused an existing function for this part

src/search.js
zanderle: The reason I used '_` is because local storage allows you to do:

`localStorage.readthedocs_search_recent_searches = 'some-value'`

It's not a feature we need, but that was the reasoning, so normally you'd either do `_` or just use camelCase. But I'm fine leaving it in kebab-case

humitos: ah, OK. So, using camelCase makes more sense I' say to keep the pattern we are using on the other variables and stuffs

humitos (approved):
I think this is going in the correct direction here.

I brought in the changes from `main` and pushed a commit to format the code and build the assets to make those checks to pass.

I'd like to know @agjohnson thoughts about moving forward with this implementation. I'm happy to merge it as-is and work on styling and adding those features we mentioned in following up PRs.

zanderle:
@humitos it is done, but just to make things easier for me (not having to worry about existing structure) when I implement the clear feature. I'm already 50% done, so it shouldn't be long.

agjohnson:
How are you all testing this? I tried on https://readthedocs-client--244.org.readthedocs.build/en/244/ but I assume that shouldn't work at all

zanderle:
@agjohnson I've been testing locally. I'm not sure of other options, but I'm sure @humitos can help

@humitos ok I added the clear search functionality (it took me forever because svg in the button was capturing click events for some reason, so nothing worked and I was stuck for a while and had no idea what was going on 😆 ), so I think this can now be merged as the first iteration

humitos:
@zanderle looks good to me 👍🏼

humitos:
@agjohnson 

> How are you all testing this?

You can either:

1. Use `npm run dev` and go to `http://localhost:8000/`, which used hardcoded search results no matter what you type
2. Tell wrangler to use this version by editing https://github.com/readthedocs/common/blob/760ffb301809d614db6531ca93a08c4104f766b8/dockerfiles/force-readthedocs-addons.js#L17-L18 to use `http://localhost:8000/readthedocs-addons.js` and then use `npm run dev`  plus `inv docker.up` to hit any documentation you have built in your development instance with addons enabled.

> I tried on [readthedocs-client--244.org.readthedocs.build/en/244](https://readthedocs-client--244.org.readthedocs.build/en/244/) but I assume that shouldn't work at all

This won't work because the js file is injected by Cloudflare --and it uses the production version-- not by this repository.

src/search.js
```diff
@@ -284,6 +291,101 @@ export class SearchElement extends LitElement {
     `;
   }
 
+  renderRecentSearches() {
+    const recentSearches = this.getRecentSearches();
+    if (!recentSearches || !recentSearches.length) {
+      return html`<p>No Recent Searches</p>`;
```
agjohnson, line 297: Unless this is to match for consistency, sentence case is a good fit here.

```suggestion
      return html`<p>No recent searches</p>`;
```

src/search.js
```diff
@@ -284,6 +291,101 @@ export class SearchElement extends LitElement {
...
+    const recentSearchesString = localStorage.getItem(
+      this.recentSearchesLocalStorageKey,
+    );
+    return recentSearchesString ? JSON.parse(recentSearchesString) : [];
+  }
+
+  storeRecentSearch(block, result) {
+    const recentSearches = this.getRecentSearches().filter((recentSearch) => {
+      const b = recentSearch.block;
+      const r = recentSearch.result;
```
agjohnson, line 350: Not a big deal, but we normally use a more verbose var name in favor of a short var name.

agjohnson (approved):
This looks good to merge, I noted some very minor issues that we can address later. I didn't got through the set up to test this locally though, it was a bit more than I had time for. But this seems tested enough already :+1:

humitos:
Thanks for the work and the review here. I updated this PR with the latest changes and I will merge it once checks pass 🚀

PR #251, opened 2024-02-14
Notifications: don't go to top page when closing it

### LINKED ISSUE #246: Dismissing PR preview banner loses position on the page
## Details

* Read the Docs project URL: https://readthedocs.org/projects/pep-previews/
* Build URL (if applicable): https://readthedocs.org/projects/pep-previews/builds/23392643/

1. Go to https://pep-previews--3649.org.readthedocs.build/pep-0742/#type-narrowing-behavior

You are taken to the desired page and anchor, but a banner is displayed with the text

> ## This page was created from a pull request build
> See the [build's detail page](https://readthedocs.org/projects/pep-previews/builds/23392321/) or [pull request readthedocs/readthedocs.org#3649](https://github.com/python/peps/pull/3649) for more information.

2. Click the 'x' to dismiss the banner

## Expected Result

Stay in the same position on the page.

## Actual Result

Takes you to the top of the page.

## See also

Originally reported at https://github.com/python/peps/issues/3657

humitos:
* When clicking on the X, now it calls `preventDefault()` to avoid moving to the top
* I added some `id=` attributes in the index page to test this manually

Closes #246


agjohnson (approved):
This makes sense. I assume this is because the link has a `href="#"` or similar, so this is a good way to avoid the default behavior here.

agjohnson:
I would say it's fine as is, your fix does solve the problem. In places where we need to worry about users with JS disabled, it's best to make this a valid URL. But obviously we don't have that problem with Addons.

PR #264, opened 2024-03-04
Docdiff: allow users to define a custom `rootSelector`

### LINKED ISSUE #263: Docdiff: should also respect the `<main>` HTML tag
Right now, from checking the source code, the DocDiff addon just checks for the `role=main` query selector, but it would be cool to also check for a `<main>` HTML tag alternatively, or to be able to configure the main container query selector for the addon.

humitos:
Closes #263
Related https://github.com/readthedocs/readthedocs.org/pull/11181


src/docdiff.js
```diff
@@ -125,11 +123,9 @@ export class DocDiffElement extends LitElement {
         const parser = new DOMParser();
         const html_document = parser.parseFromString(text, "text/html");
         const old_body = html_document.documentElement.querySelector(
-          this.config.addons.doc_diff.root_selector,
-        );
-        const new_body = document.querySelector(
-          this.config.addons.doc_diff.root_selector,
+          this.rootSelector,
```
agjohnson, line 126: Given this is user configurable, we should catch errors here -- ie, for invalid selectors:

```
> document.querySelector("!!!");
< Uncaught DOMException: Document.querySelector: '!!!' is not a valid selector
```

agjohnson (approved):

src/docdiff.js
humitos: I checked this and I found this is already inside a `.then` and `.catch` async block.

https://github.com/readthedocs/readthedocs-client/blob/0995c065b45a02b206937ce2bcc6ef0973ab2b10/src/docdiff.js#L122-L141

I understand there is no need to add another try/catch here. Is it OK?

agjohnson: Ah yes, exceptions in `then` will bubble up to `catch()` here, that indeed should be enough.

PR #265, opened 2024-03-05
Notifications: respect `project.versioning_scheme` for links

### LINKED ISSUE #243: Bug: Stable version link in banner is hardcoded
We have multi-version without languages set up so our stable version link does not contain `/en`.

However, the banner that points to stable version in the latest pages has the language hardcoded... 
image
https://github.com/readthedocs/addons/blob/main/src/notification.js#L132

humitos:
This is a workaround for now since the backend API is not returning this URLs due to https://github.com/readthedocs/readthedocs-ops/issues/1323. However, we cannot fix that quickly.

I'm re-implementing a small version of the resolver for this by generating the URL based on the `project.versioning_scheme` for now.

We should come back to this once the API returns the URLs we need.

Closes #243


src/notification.js
```diff
@@ -127,9 +130,22 @@ export class NotificationElement extends LitElement {
 
     if (stable_index !== -1) {
       this.stableVersionAvailable = true;
-      // TODO: we need to use, somehow, the "resolver.resolve" logic from the Python backend
+      // TODO: use the URL comming from the backend under `addons.versions.stable.urls.documentation` or similar
+      // We need to use, somehow, the "resolver.resolve" logic from the Python backend
       // to support all the posibilities. Those cases won't work for now until we find a proper solution.
-      this.urls.stable = `/${current_project.language.code}/stable/`;
+      // For now, we are poorly re-implementing this in the frontend as a workaround, but this is not ideal.
+      // See https://github.com/readthedocs/readthedocs-ops/issues/1323
```
stsewd, line 137: The issue at ops is about not being able to resolve URLs from the dashboard, documentation URLs can be resolved from any service, webs or proxito.

humitos: You are right about this. We can resolve the documentation URL for the version from El Proxito. _However_, we would need to do a hack for this and break the pattern we are following here: re-use the API serializers and return full/completely objects.

We are currently removing `_links` and `url` attributes from each of the resources because of the linked issue:

https://github.com/readthedocs/readthedocs.org/blob/8fd472aeae0c208758d90c8d8d12eadd721f8f29/readthedocs/proxito/views/hosting.py#L182-L219

That `url` field uses `VersionURLsSerializer` which has the `documentation` URL we need here:

https://github.com/readthedocs/readthedocs.org/blob/8fd472aeae0c208758d90c8d8d12eadd721f8f29/readthedocs/api/v3/serializers.py#L319-L330

I prefer to build the URL from the JavaScript code for now until we solve the underlying backend issue and return these resources with all the expected fields.

humitos: Ideally, the fontend should use:

```javascript
const url = this.config.versions.stable.urls.documentation
```

without generating the URL here and dealing with `versioning_scheme` as I'm doing now, but we are not there yet, unfortunately.

src/notification.js
```diff
@@ -127,9 +130,22 @@ export class NotificationElement extends LitElement {
...
+      if (
+        current_project.versioning_scheme ==
+        "multiple_versions_with_translations"
+      ) {
+        this.urls.stable = `/${current_project.language.code}/stable/`;
```
agjohnson, line 142: I can't tell quickly, is this supposed to support subproject URLs or custom urlconf? It seems like these might be missed here by using the root path relative URL.

src/notification.js
stsewd: I think we should just re-add that single URL field in the overridden serializer instead of doing a big change in our codebase.

src/notification.js
stsewd: yeah, I suggest we use our resolver instead https://github.com/readthedocs/addons/pull/265#discussion_r1514938912

humitos:
I updated this PR to use the URLs returned from the backend. It requires:

* https://github.com/readthedocs/readthedocs.org/pull/11227
* https://github.com/readthedocs/readthedocs.org/pull/11205

agjohnson (approved):

humitos:
The API backend changes were merged, this PR is unblocked now and needs a small updates to make the tests pass and other small bits.

PR #268, opened 2024-03-05
Search: close the modal when clicking on a result

### LINKED ISSUE #247: Search: close modal if the resulting link is on the same page
### Steps to reproduce

1. go to https://docs.pypi.org/organization-accounts/actions/billing-actions/
2. open search modal
3. search for `billing`
4. click on the second result

### Expected result

Go to the exact anchor link and close the modal.

### Actual result

Go to the exact anchor link but keeps the modal opened.

humitos:
Clicking on a result close the search modal if the resulting link is on the same page the user is currently viewing.

Closes #247


src/search.js
```diff
@@ -249,6 +253,12 @@ export class SearchElement extends LitElement {
     `;
   }
 
+  followResultLink(e) {
+    // Close the modal if the link is on the same page
+    const event = new CustomEvent(EVENT_READTHEDOCS_SEARCH_HIDE);
+    document.dispatchEvent(event);
```
agjohnson, line 259: It's not a problem with this exact use case, but if anyone tries to do anything async from the `SEARCH_HIDE` event, the code will likely be cut off. The operation order can be something like:

- User clicks results
- Event SEARCH_HIDE is fired off
- Browser navigates to link `href`
- Event SEARCH_HIDE is never consumed, or async code like `fetch` never finishes.

The user doesn't have the ability to affect the event propagation, so the browser will always follow the link.

I'm just reading into this code, I don't have immediate suggestions. If we are advertising this event to users, it's at least worth a note. This might be a pattern to change later too though.

agjohnson (approved):

src/search.js
humitos: Good point here 👍🏼 . I thought a little about what you are describing and I'm not sure what would be good solution for this case. I used this pattern here because I want to standardize how we open/close or enable/disable addons, and I found that using custom events was a good pattern to follow here.

Besides, as you already guessed, users will be able to use the exact same pattern when integrating the addons in their theme or similar.

I will open an issue to think more about this use case, and see where we arrive here.

humitos: https://github.com/readthedocs/addons/issues/271

PR #281, opened 2024-03-26
API: simplify and standardize data usage

### LINKED ISSUE #132: Notification: handle `stable`/`latest` aliases
Currently, our logic to show a notification is:

- on `latest` show a warning about potentially unreleased features
- on non-`stable` show a warning about maybe reading an outdated documentation

We need to handle the _real_ branch/tag names these `stable` / `latest` versions refer to. Why is this? Because we want to show the `latest` notification on `main` too, for example. Also, we don't want to show the non-`stable` notification if the `stable` version points to `v3.1` and the user is reading the `v3.1`.

The API should return a list of these alias, like:

```javascript
{
  "latest": ["latest", "main"],
  "stable": ["stable", "v3.1],
}
```

and the JS code should check the current version against those.

Related #125

humitos:
Implements the changes in the API made in https://github.com/readthedocs/readthedocs.org/pull/11205. We need to merge that PR and deploy addons together with the changes in this PR.

In this PR there are mainly no changes in the logic. The most important thing to mention here is that we are standardizing the usage of the API response and pinning to `v1` now (we were using `v0-alpha`). This means:

- we will be exposing the `v1` to users via the js custom event (see #64)
- we are using standard/shared APIv3 serializers for resources for this `/addons/` endpoint
- any breaking change we have to do in the future it will increase the API response version

I'm not sure there are too much to review here, but I'd appreciate at least a quick look at these changes. I'm happy to answer some questions you may have here or on the underlying PR about the API response changes.

### Related:
* Closes #132 
* Unblocks #265
* Unblocks theme integration (our theme and CPython's one)
* Requires https://github.com/readthedocs/readthedocs.org/pull/11205


tests/__snapshots__/flyout.test.snap.js
```diff
@@ -51,8 +58,8 @@ snapshots["Flyout addon snapshot flyout completely"] =
         Downloads
       </dt>
       <dd>
-        <a href="/_/downloads/en/latest/">
-          PDF
+        <a href="https://project.readthedocs.io/_/downloads/en/latest/pdf/">
+          pdf
```
agjohnson, line 62: Oh hrm, I didn't think of this. Just reusing the attribute names we'll be dropping the casing on these options. It might be worth a lookup mapping to get this string with the same casing on these options.

As lower case, screen readers might try to pronounce `pdf` as a word instead of an acronym.

src/data-validation.js
```diff
@@ -181,17 +177,61 @@ const addons_flyout = {
             },
           },
         },
+        translations: {
+          type: "array",
+          // TODO: validate each item of the array has the following structure
```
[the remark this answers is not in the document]
humitos: 👍🏼 -- I created an issue for this https://github.com/readthedocs/addons/issues/288

humitos:
> but trust our tests would pick up issues there anyways.

I can confirm this. I took me a lot of time to adapt them to the new schema 😅

src/flyout.js
```diff
@@ -174,30 +167,35 @@ export class FlyoutElement extends LitElement {
...
 
   renderDownloads() {
-    if (!this.config.addons.flyout.downloads.length) {
+    if (!Object.keys(this.config.versions.current.downloads).length) {
       return nothing;
     }
 
+    const nameDisplay = {
+      pdf: "PDF",
+      epub: "ePUB",
```
agjohnson, line 189: Maybe either are fine, but [the specification](https://www.w3.org/AudioVideo/ebook/) uses this:

```suggestion
      epub: "EPUB",
```

agjohnson (approved):

PR #282, opened 2024-03-29
Customization via CSS variables

### LINKED ISSUE #197: Customization via CSS variables
I've been using `--readthedocs-flyout-font-size` and `--readthedocs-notification-font-size` and I found them pretty good. Mainly the ones for flyout since by defining one variable the rest of font-sizes are calculated based on only that variable. This is thank to the PR #181

However, we may need to calculate other variables based on a global setting. For example `line-height` and similar stuffs. Otherwise, when reducing the `font-size` it looks weird since the `line-height` is not reduced:

### Normal style

Screenshot_2023-11-21_16-58-04

### Using `--readthedocs-notification-font-size: 0.8rem`

Screenshot_2023-11-21_16-58-28

----

It's clear how the `line-height` is bigger in the Material for MkDocs example.

Questions:

* Is there a simple way to "change all them together" that makes sense?
* Should we add more CSS variables to change `line-height` as well? (e.g. `--readthedocs-flyout-line-height`)
* What approach should we take here?

zanderle:
This PR makes the notifications a bit more reliant on the CSS variables. I tried doing the same for search as well, but focused mostly on colors. Making the full design rely on `font-size` is a bit more tricky for the search modal, because there's more involved than the "simple box" that are the notifications. We can always improve as we move further.

I also had to change the setting of the [default CSS variable on flyout](https://github.com/readthedocs/addons/blob/9ae69be90f49096b3a975238ca1e67a78055d456/src/flyout.css#L32) from `:root` to `:host` as the former didn't seem to work.

Closes #197


humitos:
> Making the full design rely on `font-size` is a bit more tricky for the search modal, because there's more involved than the "simple box" that are the notifications

Should we have a specific variable for this? I don't know, maybe `readthedocs-<addons>-size: 100%` where all the size/spacing/etc are calculated based on that variable?

zanderle:
> > Making the full design rely on `font-size` is a bit more tricky for the search modal, because there's more involved than the "simple box" that are the notifications
> 
> Should we have a specific variable for this? I don't know, maybe `readthedocs-<addons>-size: 100%` where all the size/spacing/etc are calculated based on that variable? Would a variable like this make this easier?

That's a good idea. But no, it's more a matter of going through every single design element in the search modal, determining how it should relate to the font-size (or some kind of a global variable) and making sure it works well with it. It's just a time-consuming process more than anything else.
If you think it's an important feature to have, I can still add that.

agjohnson (approved):
All of this seems good. It is indeed a shame there isn't a way to reference the root shadow dom element using an absolute unit like `rem`.

humitos:
I will try to test this locally soon and merge after that.

humitos:
@zanderle @agjohnson how are you testing this? I added the following CSS and I'm not seeing any change:

```css
:host {
  --readthedocs-flyout-font-size: 0.2rem;
  --readthedocs-notification-font-size: 0.2rem;
  --readthedocs-search-font-size: 0.2rem;
}
```

agjohnson:
Did you try setting those variables on `<html>` or something top level and outside the shadow DOM?

humitos:
Using `html` seems to work:

```css
html {
--readthedocs-flyout-font-size: 0.63rem;  /* Material for MkDocs */
}
```

Is this the correct way we want users to customize the sizes?

One thing I noticed here is the branch icon doesn't scale accordingly with this setting:

Screenshot_2024-04-17_12-16-06


I found this because the size of this icon is bigger on Material for MkDocs than on Sphinx by default (same as the font), but changing the `--readthedocs-flyout-font-size` doesn't make the icon to scale accordingly. Is it possible to consider this as well?

zanderle:
> Using `html` seems to work:
> 
> ```css
> html {
> --readthedocs-flyout-font-size: 0.63rem;  /* Material for MkDocs */
> }
> ```
> 
> @agjohnson Is this the correct way we want users to customize the sizes?

You can also set it on `body` or on the web-component itself, and it should work the same.

> 
> One thing I noticed here is the branch icon doesn't scale accordingly with this setting:
> 
> Screenshot_2024-04-17_12-16-06
> 
> I found this because the size of this icon is bigger on Material for MkDocs than on Sphinx by default (same as the font), but changing the `--readthedocs-flyout-font-size` doesn't make the icon to scale accordingly. @zanderle Is it possible to consider this as well?

Yes, I'll add that. Good catch.

humitos:
@zanderle can you tell me what are the steps you are following to test this? I'm having issues with. I'm not sure why sometimes it works and sometimes it doesn't. I may not be following the correct steps here.

zanderle:
> @zanderle can you tell me what are the steps you are following to test this? I'm having issues with. I'm not sure why sometimes it works and sometimes it doesn't. I may not be following the correct steps here.

When I'm testing, I do it in developer tools, and create the variable on the element itself:

Screenshot 2024-04-17 at 16 17 46

humitos:
OK, yeah. Testing it that way it works for me as well (and the SVG icon scales properly 👍🏼 ). However, it's still not clear to me how users should define these variables. We talked about defining them under `:host`, `:root`, `html` and/or `body`. However, none of them work for me now. @zanderle do they work for you?

@agjohnson how do we want users to define these variables? Can you provide an example about this?

humitos:
> I tried doing the same for search as well, but focused mostly on colors. Making the full design rely on `font-size` is a bit more tricky for the search modal, because there's more involved than the "simple box" that are the notifications. We can always improve as we move further.

I opened https://github.com/readthedocs/addons/issues/292 to track and work on this in a following PR 👍🏼

zanderle:
> OK, yeah. Testing it that way it works for me as well (and the SVG icon scales properly 👍🏼 ). However, it's still not clear to me how users should define these variables. We talked about defining them under `:host`, `:root`, `html` and/or `body`. However, none of them work for me now. @zanderle do they work for you?

Yeah, I'm noticing it now too. Not sure about the "sometimes it works, sometimes it doesn't", but it looks like the css variable defined in the web-component using `:host { }` takes precedence over anything defined in the parent DOM (even with `!important`), since that's the scope it's in. However if we don't define it using `:host { }` we can't provide the default value for the variable.
The way around this is that we just have to define it in a way that's more specific than `:host`. So in the parent DOM we could either have

```css
/* Using the element */
readthedocs-flyout {
  --readthedocs-flyout-font-size: 1.2rem;
}
```

or if we defined the webcomponent to have a specific class, we could do:

```
/* Using the class */
.some-specific-class {
  --readthedocs-flyout-font-size: 1.2rem;
}

...

<readthedocs-flyout class="some-specific-class"></readthedocs-flyout>
```

If that's not acceptable, I'll have to do some more research to see if I'm missing something

agjohnson:
So, all of these options should operate the same, they are all valid methods.

However, our variables are scoped per-addon (`--readthedocs-flyout-font-size`) so I think our guidance should be setting the variables at the top level. At the parent, the `html` or `:root` selector are most common.

> However if we don't define it using :host { } we can't provide the default value for the variable.

There are two ways I see us getting around this:

First, we can use an internal variable name that doesn't conflict with the outer variable name. Inside the flyout CSS for example:

```css
:host {
  --readthedocs-inner-flyout-font-size: var(--readthedocs-flyout-font-size, 0.8rem);
}
```

This seems easy and low friction, I didn't test this though and maybe I'm wrong here.

----

Maybe a larger wrench in our plans than we wanted is a different version of the above though.

At the parent level, users could do any of these:

```css
:root {
  --readthedocs-font-size: 19px;
}

readthedocs-flyout {
  --readthedocs-font-size: 20px;
}

readthedocs-flyout.bigger {
  --readthedocs-flyout-font-size: 21px;
}
```

And our addons set the variables as:

```css
:host {
  --readthedocs-flyout-font-size: var(--readthedocs-font-size, 0.8rem);
}

p {
  font-size: var(--readthedocs-flyout-font-size);
}
```

Giving us shared CSS variables at `html`/`:root`/whatever and individual addons variables at the addon element specificity.

zanderle:
> There are two ways I see us getting around this:
> 
> First, we can use an internal variable name that doesn't conflict with the outer variable name. Inside the flyout CSS for example:
> 
> ```css
> :host {
>   --readthedocs-inner-flyout-font-size: var(--readthedocs-flyout-font-size, 0.8rem);
> }
> ```
> 
> This seems easy and low friction, I didn't test this though and maybe I'm wrong here.
> 

Yeah, this one should be pretty easy to do - it would just be a matter of renaming the "inner" variables.

> Maybe a larger wrench in our plans than we wanted is a different version of the above though.
> 
> At the parent level, users could do any of these:
> 
> ```css
> :root {
>   --readthedocs-font-size: 19px;
> }
> 
> readthedocs-flyout {
>   --readthedocs-font-size: 20px;
> }
> 
> readthedocs-flyout.bigger {
>   --readthedocs-flyout-font-size: 21px;
> }
> ```
> 
> And our addons set the variables as:
> 
> ```css
> :host {
>   --readthedocs-flyout-font-size: var(--readthedocs-font-size, 0.8rem);
> }
> 
> p {
>   font-size: var(--readthedocs-flyout-font-size);
> }
> ```
> 
> Giving us shared CSS variables at `html`/`:root`/whatever and individual addons variables at the addon element specificity. Or combine the two so that users aren't thrown off setting `--readthedocs-flyout-font-size` at `:root`.

Yeah, I think the combination would be a good approach too. I'll try putting this together to see if there are any hidden problems and then we can decide on the final "format".

zanderle:
I wasn't sure about the naming, but take a look at the last commit in this PR. How do you feel about this approach? @humitos @agjohnson

humitos:
> I wasn't sure about the naming, but take a look at the last commit in this PR. How do you feel about this approach? 

Looks good to me 👍🏼

agjohnson:
It looks like `--addons-*` is the internal variable name that we manage, and `--readthedocs-*` is the external, user controlled variable name?

@humitos you are generally a proponent of keeping the `--readthedocs-*` prefix the same on all variables, does having the two competing prefixes seem okay to you?

Otherwise this approach looks great to me.

humitos:
I'm not too worried about the _internal_ name. We can always change it later if we find a better one. However, I want to keep `--readthedocs-*` prefix for all variables we are exposing to users.

zanderle:
@humitos should be ready to be merged

humitos:
I think this is fine as a first step. However, I don't feel super comfortable with it yet and I'm not sure if it will block us to make changes in the future to our addons without breaking people's customization. Fortunately, we will have the customization scoped by the variables we are exposing here but I'm still a little worried --mainly because of my lack of CSS knowledge here.

I tested this PR and it seems it works better to what we currently have, so I'm happy to merge. However, there are still some improvements we need --mainly in the search modal since defining these variables doesn't apply to some elements (eg. "Filter" subtitle and "Include subprojects")

Screenshot_2024-06-18_16-10-35

Thansk @zanderle for this work!

zanderle:
@humitos maybe we can mitigate this by making it clear in the docs that things might change?

humitos:
@zanderle yes -- we don't even have docs for this yet 😅 , but we should mention that once we write them.

PR #286, opened 2024-04-09
Refactor `getReadTheDocsConfig` to use `Promise`

humitos:
Follow Anthony's suggestion from
https://github.com/readthedocs/addons/pull/64#discussion_r1361023485


src/readthedocs-config.js
```diff
@@ -59,63 +61,80 @@ export function getReadTheDocsConfig(sendUrlParam) {
...
+          reject("Error hitting addons API endpoint");
+        }
+        // Use the addons API data response as `dataUser`
+        dataUser = response.json();
+        return dataUser;
+      })
+      .then((data) => {
+        // Create a new Promise here to handle the user request in a different async task.
+        // This allows us to start executing our integration independently from the use one.
+        new Promise((resolve, reject) => {
```
agjohnson, line 90: Is this `Promise` instance needed? You are already in an async function at this point, and should be able to chain into the next `fetch` call from what I can tell.

humitos: I did this on purpose to create a different task so it gets executed async and don't block the current one. This is to avoid hitting the API twice _before_ executing our own addons. Does this work as I expect?

agjohnson: Ah gotcha, yeah I think I see what you are going for here now. This seems good then, this is the right approach.

agjohnson: I don't think it should matter much, but the argument names here are overlapping with the parent Promise. To avoid unexpected errors, these could be explicit:

```suggestion
        new Promise((resolveUser, rejectUser) => {
```

src/readthedocs-config.js
```diff
@@ -59,63 +61,80 @@ export function getReadTheDocsConfig(sendUrlParam) {
...
+                "client-version": CLIENT_VERSION,
+                "api-version": metadataAddonsAPIVersion,
+              });
+
+            fetch(userApiUrl, {
+              method: "GET",
+            }).then((response) => {
               if (!response.ok) {
-                throw "Error parsing configuration data";
+                reject(
```
agjohnson, line 115: One thing to be careful of with using `Promise` is to make sure you are using `return` to end the function on resolve/reject, otherwise this will continue on in the promise chain.

```suggestion
                return reject(
```

src/readthedocs-config.js
```diff
@@ -59,63 +61,80 @@ export function getReadTheDocsConfig(sendUrlParam) {
...
-          url =
-            ADDONS_API_ENDPOINT +
-            new URLSearchParams({
-              url: window.location.href,
-              "client-version": CLIENT_VERSION,
-              "api-version": metadataAddonsAPIVersion,
-            });
+      .then((response) => {
+        if (!response.ok) {
+          reject("Error hitting addons API endpoint");
```
agjohnson, line 81: Noted below, but `reject` doesn't end the function by itself, so this will actually continue on to the `response.json()` below. Usually you add a `return` to force the end of the promise:

```suggestion
          return reject("Error hitting addons API endpoint");
```

src/readthedocs-config.js
```diff
@@ -59,63 +61,80 @@ export function getReadTheDocsConfig(sendUrlParam) {
...
-                EVENT_READTHEDOCS_ADDONS_DATA_READY,
-                document,
-                data,
-              );
-            })
-            .catch((error) => {
-              console.error(error);
+              // If the user defined a meta HTML tag with a different api-version,
+              // use the new API data response as `dataUser`
+              dataUser = response.json();
```
agjohnson, line 121: I noted this in another PR, but I think we should resolve this promise instead of passing the promise on to the user. I can't do a suggestion here, but loosely what I mean is:

```js
fetch(...).then((response) => {
  ...
  return response.json();
}).then((data) => {
  dispatchEvent(..., data);
});
```

This is because we are already firing off the event as an async event, and so it feels odd to then make the user await on the data in addition -- we can do both at once.

agjohnson: I noted similar here: https://github.com/readthedocs/addons/pull/287#discussion_r1558299779

The example above is for illustration, I might have the promise execution order confused a bit.

src/readthedocs-config.js
```diff
@@ -64,73 +67,80 @@ function _getApiUrl(sendUrlParam) {
   return url;
 }
 
+function getReadTheDocsUserConfig(sendUrlParam) {
```
agjohnson, line 70: Breaking this out is a good idea :+1: Often, splitting up nested async/promises makes the code easier to follow

agjohnson (approved):
This looks good, the logic matches what I would expect :+1:

PR #287, opened 2024-04-09
CustomEvent: pass a `ReadTheDocsData` object as `event.detail`

### LINKED ISSUE #279: CustomEvent: communicate subcribed to event without `meta` tag defined
It would be good if we can communicate to the user they are subscribed to the `readthedocs-addons-data-ready` but they haven't defined the `meta[name=readthedocs-addons-api-version]` tag which is mandatory to make usage of the Read the Docs data event.

Reference: https://github.com/readthedocs/addons/pull/64/files#r1331997049

humitos:
I found a pattern that I like 👍🏼 . I think it's a clear way to communicate users what's going on, but also leaves the door open to expand checks/validations in the future. I also imagine we could even want to know "What projects are subscribing to the custom event?" in case we need to contact them due to a breaking change or similar.

#### Calling `event.detail.initialize()` without having defined the `<meta ... />` HTML tag:

Screenshot_2024-04-09_13-57-20

#### Calling `event.detail.data()` without calling `.initialize()` first:

Screenshot_2024-04-09_13-56-01

Closes #279


src/events.js
```diff
@@ -7,3 +10,49 @@ export const EVENT_READTHEDOCS_FLYOUT_SHOW = "readthedocs-flyout-show";
...
+ *     function (event) {
+ *       // Initialize the class first
+ *       event.detail.initialize();
+ *
+ *       // Use the data when it's ready
+ *       event.detail.data().then((data) => {
+ *         console.log(data.projects.current.slug);
+ *       });
+ *     }
+ *   );
```
agjohnson, line 32: I feel like when we trigger this event -- `data-ready` -- the data should be ready, and the underlying promise from `data()` should be resolved at that point.

It also feels odd to require two method calls, it seems like we can initialize and return the resolved data in one function or method.

That is:

```suggestion
 *   document.addEventListener(
 *     "readthedocs-addons-data-ready",
 *     function (event) {
 *       const data = event.detail.data();
 *     }
 *   );
```

This feels more expected.

src/events.js
```diff
@@ -7,3 +10,49 @@ export const EVENT_READTHEDOCS_FLYOUT_SHOW = "readthedocs-flyout-show";
...
+
+    this._initialized = true;
+  }
+
+  data() {
+    if (this._initialized) {
+      return this._data;
+    }
+    throw "ReadTheDocsData hasn't been initialized yet. Calling '.initialize()' is required to use the '.data' property.";
+  }
```
agjohnson, line 57: From above, I meant:

```suggestion
  data() {
    if (!this._initialized) {
      this.initialize();
    }
    return this._data;
  }
```

src/readthedocs-config.js
```diff
@@ -126,7 +128,7 @@ export function getReadTheDocsConfig(sendUrlParam) {
           dispatchEvent(
             EVENT_READTHEDOCS_ADDONS_DATA_READY,
             document,
-            dataUser,
+            new ReadTheDocsEventData(dataUser),
```
agjohnson, line 131: And in terms of resolving the promise before calling the event, `dispatchEvent` would trigger inside a promise `then()`.

agjohnson (commented):
I noted how this pattern could be improved for the user. I think we can also just use a function for this structure, it's not adding too much.

humitos:
> The class structure isn't a problem, but could be easier as a function too, the class is maybe not adding much.

I also think the class is not adding too much right. However, do you think we will have the same flexibility with a function in the future in case we need to perform more validations/checks? I'm asking because my limited js knowledge 😅

src/readthedocs-config.js
humitos: I didn't understand your comment here. Can you expand on this? 🙏🏼

humitos: I read this comment a couple of times and I also tested this locally to find our how to resolve this Promise. I made these changes in a92994a. Let me know if that sounds good to you.

agjohnson: So what I mean is that `response.json()`/`dataUser` is a promise, and we should give the user the actual value that is provided by resolving/fulfilling `response.json()`. You seem to have it close I think, but I feel like your example might be prematurely resolving the parent Promise perhaps. Let me digest this one a bit more.

agjohnson: Also, see https://github.com/readthedocs/addons/pull/286#discussion_r1559797056

agjohnson: I just commented on the commit directly https://github.com/readthedocs/addons/commit/a92994a6a9fdff536b28ff1182341fe0e0128c9c

But to keep conversation here, you may have resolved the wrong promises (both the API promises and the user JSON data promises).

agjohnson:
> However, do you think we will have the same flexibility with a function in the future in case we need to perform more validations/checks? I'm asking because my limited js knowledge

We would have plenty of flexibility to include logic that we want to hide from the maintainer user. But, if we wanted to expose and public functions/methods, the class structure would make this cleaner. I think it's fine to stick with the class based approach if we aren't sure or are leaning towards exposing public methods -- ie, maybe a `event.detail.getVersions()` helper method or something similar.

humitos:
It seems we both agree on passing a class instead of the raw JSON data. So, I'm merging this PR into the one its based, and we can continue with the minor details there.

PR #291, opened 2024-04-17
API: match changes in response, `versions.actives`

humitos:
Follows https://github.com/readthedocs/readthedocs.org/pull/11286


agjohnson (commented):
Just noting here again, but `active` is fine as it's an adjective.

PR #295, opened 2024-04-24
EthicalAd: use `placement-bottom` to show the ad above the flyout

humitos:
This PR enables showing EthicalAds for all projects hosted on Read the Docs and using the new beta addons.

Requires https://github.com/readthedocs/ethical-ad-client/issues/188


src/ethicalads.js
```diff
@@ -71,37 +71,34 @@ export class EthicalAdsAddon extends AddonBase {
...
+        placement.setAttribute(
+          "data-ea-campaign-types",
+          campaign_types.join("|"),
+        );
+      }
+
+      placement.classList.add("raised");
+
+      // TODO: find the right last resourse to append (probably not `document.body`)
+      let main = document.querySelector("[role=main]") || document.body;
```
ericholscher, line 98: I think we've written this logic in a few places -- we should standardize on it.

Also should we bring over any of the logic around theme placements (https://github.com/readthedocs/readthedocs.org/blob/main/readthedocs/core/static-src/core/js/doc-embed/rtd-data.js#L10-L18)? It's pretty Sphinx specific, but probably still useful to have the ability to smartly place things. 

Looking at the old code though, it looks like we're only injecting the ad if we're confident that we know what theme it is, so this is probably a better place to start, and not try to make it too smart and see how it works :)

ericholscher (approved):

src/ethicalads.js
ericholscher: Thinking something like:
```
Placement preference for RTD ad injection:

Using stickybox image (These float if the screen is wide, automatically):
    * Sidebar like content (`role=sidebar`)
        (Currently we check for sidebar height, and if it's too long, we put it in the footer -- but this is probably not necessary?)
    * Footer like content (`role=footer`)
    * Main body content (`role=main`)
Using text:
    * Fixed footer (Not conflicting with the flyout)

```

humitos: I opened https://github.com/readthedocs/addons/pull/309 with the logic from the old implementation.

davidfischer: As to finding the main div, this is what the ad analyzer does. It follows [these](https://github.com/readthedocs/ethical-ad-server/blob/9447f5b9df15ae98e7c117aa313da3358ee2377d/adserver/analyzer/backends/base.py#L19-L23) in order.

As Eric mentioned, we styled the ad based on the theme if we were confident in the theme being RTD-like or Alabaster-like (including RTD-like on mkdocs). This is mostly because Alabaster is "light" while RTD theme is "dark". The current ad client has light and dark theme and that could be used instead if desired. It might allow us to remove some ads-specific code in the RTD codebase. The end goal is just to have the ad look good on a variety of themes regardless of how that actually happens.

PR #299, opened 2024-05-01
Fix the hotkeys after search modal has been opened and closed

### LINKED ISSUE #169: Search: doesn't open after closing by clicking outside
Steps to reproduce:

- hit `/` to open the search modal
- click outside the modal to close it
- hit `/` again
- it does not open as expected

This is because the state saved inside the `hotkeys.js` file is not updated when clicking outside.

### LINKED ISSUE #172: Search: close the modal with `Esc` as shown in the bottom
We have this documented at the bottom of the modal but it's not currently working.

Take into consideration #169 when working on this, since we will need to share the state between these two again.

zanderle:
Closes #169 
Closes #172


src/search.js
```diff
@@ -511,6 +514,11 @@ export class SearchElement extends LitElement {
         window.location.href = selected.href;
       }
     }
+
+    if (e.keyCode === 27) {
```
humitos, line 518:
```suggestion
    # If "Esc" key is pressed.
    if (e.keyCode === 27) {
```

humitos (commented):
Looks good to me. I will do a quick local test when I can and merge after that.

src/search.js
agjohnson: :snake:

stsewd: I think you can also do (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code, https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key)

```suggestion
    if (e.key === "Escape") {
```

keycode is deprecated.

zanderle: Oh, I did not know that. Should I change other instances of `.keyCode` in this function as well? cc @humitos

humitos: I didn't know either. Sounds good to me 👍

humitos (approved):
Works great for me! Thanks 👍🏼

PR #303, opened 2024-05-08
Flyout: remove privacy policy link

humitos:
We can add a better link once we have built the documentation for the flyout/addons.

Closes #302.

PR #312, opened 2024-05-21
Plausible: add UTM analytics parameters

### LINKED ISSUE #307: Flyout: add better `utm_` parameters to the links
We are using only `ref=` for now, but we can be a little smarter here and add `utm_source=` and `utm_content=<project-slug>` so we can understand more from what particular project the user clicked the link.

humitos:
We talked about knowing more where the users come from. We are starting by adding UTM parameters in the addons so we can filter them via Plausible.

This is a better version of https://github.com/readthedocs/addons/pull/303#discussion_r1599132115

Closes #307


src/flyout.js
```diff
@@ -194,7 +216,11 @@ export class FlyoutElement extends LitElement {
         <dt>Downloads</dt>
         ${Object.entries(this.config.versions.current.downloads).map(
           ([name, url]) =>
-            html`<dd><a href="${url}">${nameDisplay[name]}</a></dd>`,
+            html`<dd>
+              <a href="${addUtmParameters(url, "flyout")}"
+                >${nameDisplay[name]}</a
+              >
```
ericholscher, line 222: Is this normally how the `a` tag is split? Feels odd to me, but it's done everywhere 🤷

src/utils.js
```diff
@@ -136,3 +136,15 @@ export function setupLogging() {
...
+export function addUtmParameters(url, content) {
+  const metaProject = document.querySelector(
+    "meta[name='readthedocs-project-slug']",
+  );
+  const projectSlug = metaProject.content;
+  const params = {
+    utm_source: projectSlug,
+    utm_content: content,
+  };
+  return url + "?" + new URLSearchParams(params);
```
ericholscher, line 149: Should we check that the URL doesn't already have a `?`?

tests/__snapshots__/flyout.test.snap.js
```diff
@@ -58,7 +58,7 @@ snapshots["Flyout addon snapshot flyout completely"] =
         Downloads
       </dt>
       <dd>
-        <a href="https://project.readthedocs.io/_/downloads/en/latest/pdf/">
+        <a href="https://project.readthedocs.io/_/downloads/en/latest/pdf/?utm_source=project&amp;utm_content=flyout">
```
ericholscher, line 61: I'm a little confused the `&` is being escaped, but nothing else in the URL is (like `?`). Is this correct?

ericholscher (approved):

src/flyout.js
humitos: I don't like it either, but this is the default `prettier` settings. Maybe it gets confused because it's HTML inside a JS file? cc @agjohnson do you know if this is a bug or a missconfiguration?

tests/__snapshots__/flyout.test.snap.js
humitos: Yes, it's correct. Using just `&` makes the tests to fail 🤷🏼 

On the other hand, the rendered version of this in the flyout links are correct and they work properly from the browser. So, I suppose it's OK and it's just something weird on the test environment.

src/utils.js
humitos: Good point. I used `new URL` and `searchParams.append()` to handle this 👍🏼

PR #315, opened 2024-05-21
Notifications: promote beta dashboard from build details page

humitos:
The link from the pull request notification points to the new beta dashboard while we are in the transition. The idea behind this is to promote the new beta dashboard a little more and get more users using it.


src/notification.js
```diff
@@ -54,7 +54,11 @@ export class NotificationElement extends LitElement {
       this.config.versions.current.type === "external"
     ) {
       this.urls = {
-        build: config.builds.current.urls.build,
+        // NOTE: point users to the new beta dashboard for now so we promote it more.
+        // We will revert this once we are fully migrated to the new dashboard.
+        build: config.builds.current.urls.build
+          .replace("readthedocs.org", "beta.readthedocs.org")
+          .replace("readthedocs.com", "beta.readthedocs.com"),
```
agjohnson, line 61: This is a naive replacement, so could backfire if `readthedocs.org` is anywhere else in the URL besides the hostname. This is not a valid project slug, so I don't think this will happen.

agjohnson (approved):

src/notification.js
humitos: He, yeah. This is a pretty dumb replacement, but I think it's enough for now while we are in the transition. The new domain will come from the API eventually.

PR #319, opened 2024-05-23
EthicalAds: force ad mentioning "Coming soon"

### LINKED ISSUE #318: Make a plan to ship house ads first to Addons users
David mentioned there is a way to force a specific ad, we should probably use that parameter.

humitos:
Closes #318


ericholscher (approved):
Looks good 👍 Would be cool if we could control this from a feature flag, but baby steps :)

humitos:
We can release only the addons pretty quickly if we want to change the ad without doing a full deploy. It's pretty simple

PR #320, opened 2024-05-23
Flyout: promote beta dashboard from it

humitos:
Use a dumb URL replacement for now to point users to the new beta dashboard. This is temporal while we are on the transition. Eventually, this will be removed and the correct URL will come from the API once we are migrated.

Related
* #315
* https://github.com/readthedocs/readthedocs-ops/issues/1485


src/flyout.js
```diff
@@ -190,7 +194,9 @@ export class FlyoutElement extends LitElement {
         <dd>
           <a
             href="${addUtmParameters(
-              this.config.projects.current.urls.downloads,
+              this.config.projects.current.urls.downloads
+                .replace("readthedocs.org", "beta.readthedocs.org")
+                .replace("readthedocs.com", "beta.readthedocs.com"),
```
agjohnson, line 199: This view does now currently exist on the new dashboard. See https://github.com/readthedocs/ext-theme/issues/77

It isn't completely useless, but this view is not linked and was never styled:

https://beta.readthedocs.org/projects/docs/downloads/

agjohnson (commented):
I think we should just hold off on these links and wait for `app.readthedocs.org` instead of doing this change multiple times.

src/flyout.js
humitos: Do you think we should remove the "Downloads" link from the flyout?

humitos:
Next week we are deploying the `app.` domain. I'm marking this as blocked on that for now.

src/flyout.js
agjohnson: Yeah, probably should drop the link to the downloads view. The downloadable formats are already listed in the flyout already, so it does seem redundant already. If users are using the downloads link we can reconsider whether we use the view or not.

humitos: I've removed this link in https://github.com/readthedocs/addons/pull/322

humitos:
This PR is ready to be merged now that we deployed `app.` domain.

agjohnson (approved):

PR #323, opened 2024-05-28
Flyout: collapse it when clicking outside

### LINKED ISSUE #110: Flyout: collapse when clicking outside it

humitos:
We broke this behavior because we overwrote these methods when working in another issue.

Closes #110


src/flyout.js
```diff
@@ -50,21 +50,12 @@ export class FlyoutElement extends LitElement {
   }
 
   _onOutsideClick = (e) => {
+    console.log("click outside");
```
agjohnson, line 53: Assuming this can probably be removed, but at least should be:

```suggestion
    console.debug("Click outside");
```

agjohnson (approved):

src/flyout.js
humitos:
```suggestion
```

humitos: Meh, this was just debugging 😅

PR #329, opened 2024-05-29
EthicalAds: inject ad only if it's a known theme and above the fold

### LINKED ISSUE #328: EthicalAds: inject ad only if we detect a theme we know
Yesterday we deploy the initial version for EA addons that enabled it in all the projects hosted on Read the Docs. However, we rolled it back because we found some inconsistencies and we didn't feel comfortable with it.

We ended up with the following plan:

> I think the first step for next addons release is to "only inject the ad if we detect is a theme we know, avoid using stickybox" until we decide what to do here and how to solve it

Slack Discussion: https://readthedocs.slack.com/archives/C02J81RSF/p1716916646608969

### Related
* https://github.com/readthedocs/addons/issues/311

humitos:
Disable using `stickybox` style for now and only inject the ad if we are sure it's a known theme and the ad it's going to be above the fold.

In a following iteration, we will want to use `stickybox` in the other cases but we want to have the ability to disable the re-placement in smaller screens.

Closes #328


ericholscher:
#330 looks good on top of this 👍

ericholscher (approved):

PR #330, opened 2024-05-29
EthicalAds: new `stickybox` logic

humitos:
This commit implements the following logic to place an ad in order:

1. If there is an explicit placement, we use that placement.

2. If the doctool/theme is known **and** we find a right placement (eg. sidebar),
   **and** the ad is gonna be above the fold, we put the ad there since we know
   it's a good fit for it.

3. Otherwise, if the screen width is >1300px we add a floating `stickybox` ad.
   Besides, we add an event handler that will _hide the ad_ if the user resizes
   the window and it ends up being <1300px.

On top of https://github.com/readthedocs/addons/pull/329


src/ethicalads.js
```diff
@@ -324,10 +296,24 @@ export class EthicalAdsAddon extends AddonBase {
     });
   }
 
+  addWidthListener() {
+    const placement = document.querySelector("#readthedics-ea");
```
ericholscher, line 300:
```suggestion
  addWidthListener() {
    // Hide ad when the window is smaller than the stickybox width
    const placement = document.querySelector("#readthedics-ea");
```

ericholscher (approved):
Interesting approach 👍

PR #335, opened 2024-05-30
EthiccalAds: use `data-ea-publisher=readthedocs` as placement

humitos:
Add an extra selector for explicit placement. This is what the Ethical Ad client documentation says and it's what Furo is using.

Screenshot_2024-05-30_16-14-43


src/ethicalads.js
```diff
@@ -5,7 +5,10 @@ import styleSheet from "./ethicalads.css";
 import { IS_TESTING } from "./utils.js";
 
 // https://docs.readthedocs.io/en/stable/advertising/ad-customization.html#controlling-the-placement-of-an-ad
-const EXPLICIT_PLACEMENT_SELECTOR = "#ethical-ad-placement";
+const EXPLICIT_PLACEMENT_SELECTORS = [
+  "#ethical-ad-placement",
+  '[data-ea-publisher="readthedocs"]',
```
davidfischer, line 10:
```suggestion
  '[data-ea-publisher]',
```

Just a thought that this should perhaps not hardcode `readthedocs` in there in case there's another publisher using an explicit placement.

davidfischer (approved):

src/ethicalads.js
humitos: Done in [`acbb395` (#335)](https://github.com/readthedocs/addons/pull/335/commits/acbb3958b751dc74914efd89af4a6bed0619c282)

PR #337, opened 2024-06-03
Docs: add environment variable

humitos:
The theme should include this into the `html_context` automatically, but for some reason this is not working like that.


stsewd:
Seems to be under this conditional

https://github.com/pradyunsg/furo/blob/0439e84b3c2c5b168f22ea2e93c867fd37911dfa/docs/conf.py#L122-L123

PR #346, opened 2024-07-15
Flyout: add link to addons documentation

### LINKED ISSUE #304: Flyout: add a link to flyout/addons documentation
In #302 we removed the Privacy Policy link from the footer, but we didn't have a better one to put there. This issue is to discuss/find a better link to put there. I think it could be one that goes to the Read the Docs Addons documentation page where the addons are explained/documented.

Related:
* https://github.com/readthedocs/readthedocs.org/issues/11187

humitos:
Closes #304


src/flyout.js
```diff
@@ -81,6 +81,16 @@ export class FlyoutElement extends LitElement {
   renderFooter() {
     return html`
       <small>
+        <span>
+          <a
+            href="${addUtmParameters(
+              "https://docs.readthedocs.io/page/addons.html",
```
agjohnson, line 87: This seems okay for now, but I'm not sure reader users will get much value out of this documentation.

It is already seeming very much for maintainers and so feels like not a good match for reader users. The content relevant to reader users will likely need to be more of a marketing page.

src/flyout.js
```diff
@@ -81,6 +81,16 @@ export class FlyoutElement extends LitElement {
...
+              "flyout",
+            )}"
+            >Addons documentation</a
```
agjohnson, line 90: I'd ensure the long text here doesn't cause the text and the "Hosted by" to wrap multiple lines, which won't look good.

agjohnson (approved):

src/flyout.js
humitos: Agreed. However, I think we don't have a better link for now. The other option could be https://about.readthedocs.com/features/reader/ --but that doesn't explain anything really, it's pretty marketing focused 😄

src/flyout.js
humitos: It renders fine here, but I'm not sure if I should check something else?


Screenshot_2024-07-18_12-22-16



and also in a project with just 1 version

Screenshot_2024-07-18_12-23-31

src/flyout.js
agjohnson: Yup, that's why I still think it's okay for now. We should aim to replace the link with something helpful to readers.

src/flyout.js
agjohnson: The width of the text will likely control the width of the flyout it seems, so it might not wrap. That looks okay to me either way

PR #352, opened 2024-07-28
Fix search selectNextResult behaviour

### LINKED ISSUE #316: Search: arrow keys navigation is inconsistent and skips entries
The navigation by arrow keys is not what I would expect, I can't tell what the pattern to movement is actually:

[Screencast from 2024-05-21 19-53-57.webm](https://github.com/readthedocs/addons/assets/1140183/73fbc946-abbf-444b-a263-bb85a60d058d)

zanderle:
I decided to fix it by changing the logic of `selectNextResult`. I hope that's ok. It was relying on the ids being set correctly, which felt very unreliable and flaky. I was looking through git history but couldn't find good reasoning behind that logic. So I instead made it rely just on the order in which they are present in the document. That felt closer to what we actually want it to do.

Let me know if you'd rather we keep the original logic.

Fixes #316


agjohnson (approved):
This seems like a better approach :+1:

humitos (approved):
Works great to me! 💪🏼

PR #353, opened 2024-07-28
Fix recent searches when looking at different versions on same domain

### LINKED ISSUE #296: Search: "Recent searches" does not consider different versions
It seems the "Recent searches" shows all the searches performed in the domain. However, in the same domain there could be different versions and even different projects when using subprojects.

We should save the `project.slug` and `version.slug` in the recent searches data to know if they match when restoring them from the cache.

How to reproduce:

1. go to https://test-builds.readthedocs.io/en/latest/
2. search for "configuration"
3. click on the result
4. go to a different version using the flyout
5. click on the search input from the flyout

At this point, you will see the "Recent search" you perform in a different version. If you click there, you will go to the other version.

zanderle:
Fixes #296 

A few considerations here:
- I used the new localStorage code that we added in #283 . This means the key for localStorage will change once this is deployed, meaning that previous recent searches won't show up anymore (and there will be an unused object in localStorage lying around). Are we ok with that, or do we want to make it backward compatible?
- I didn't add any logic for removing recent searches from previous versions. If we don't add anything, this object will grow and grow and it will store recent searches of versions that might become very old. Do we want to add some kind of logic in this PR or should we do it in a separate PR?
- Besides project and version, I also added language to the key (the same way we do it for notifications). Is that ok?


agjohnson (approved):
This looks good to me. I don't think we need to be backwards compatible here, very likely this won't be noticed either way. I think instead of building any sort of TTL, we should just limit the size of the recent searches to 5/10/whatever results. This would effectively and naturally remove old search terms from the list without us adding date based logic.

Using language and version here is indeed the most correct, both could affect where the result links to.

zanderle:
>  I think instead of building any sort of TTL, we should just limit the size of the recent searches to 5/10/whatever results. This would effectively and naturally remove old search terms from the list without us adding date based logic.

Yeah, that was the initial implementation I had (and that's still in effect) for a **single** project/version/language. This worked fine when all we had was just 1 version per domain (the incorrect initial assumption). However now that we have multiple versions, it's not clear what the logic should be. We could timestamp when certain recent searches are created, and then start removing the oldest ones first (regardless of the version)? What do you think?

agjohnson:
Even with the new storage pattern, storing the recent searches per project + version + language, can't we still just truncate each of these arrays at 5/10/whatever search queries?

I feel a time based approach will still have edge cases and doesn't have a whole lot of opportunity to be more accurate. I suppose I'm thinking of how browsers work mostly -- there is no date based invalidation to search queries/history. I think it will be expected that our search history functions the same.

zanderle:
> Even with the new storage pattern, storing the recent searches per project + version + language, can't we still just truncate each of these arrays at 5/10/whatever search queries?

That is the implementation we have 👍 
 
> I feel a time based approach will still have edge cases and doesn't have a whole lot of opportunity to be more accurate. I suppose I'm thinking of how browsers work mostly -- there is no date based invalidation to search queries/history. I think it will be expected that our search history functions the same.

True. But because we're only limiting each of those arrays, we don't have any limit on the amount of overall searches stored. I don't think it's a huge deal, but I thought it's still worth pointing out.

agjohnson:
Gotcha, yeah if we get to storing a lot of data that is good to be mindful of. It seems like we don't have too much to store at the moment at least.

Issue #292, opened 2024-04-17
Search: scale modal based on `--readthedocs-search-font-size`


humitos:
> I tried doing the same for search as well, but focused mostly on colors. Making the full design rely on font-size is a bit more tricky for the search modal, because there's more involved than the "simple box" that are the notifications. We can always improve as we move further.

_(from https://github.com/readthedocs/addons/pull/282)_

PR #354, opened 2024-08-02
Fix the filters title covering other labels

### LINKED ISSUE #293: Search: style difference between Firefox and Chrome
We found this in our own docs: https://docs.readthedocs.io/en/latest/

Screenshot_2024-04-18_17-50-28

zanderle:
Fix #293 

There's still a slight difference. The only way to avoid that completely, would be to remove the default checkbox, and style our own. But I'm not really keen on that solution. 

BTW this styling/positioning will probably be further improved in #292 , so this is just a quick solution to make it look better.


agjohnson (approved):
This is totally reasonable. Honestly, I'm not sure these elements really need to be absolutely positioned, but you work with what you have...

Looks good :+1:

humitos:
In my tests, the checkbox is still covered by the "Filters" border/padding or similar in Chrome. I'm having the same behavior reported in the issue. @zanderle can you share a screenshot on how it behaves for you?

humitos:
@zanderle were you able to take a look at this?

zanderle:
Should be ok now. Sorry for the delay.

Screenshot 2024-10-18 at 11 23 32

Issue #357, opened 2024-08-06
Search: we are showing `true` in the result list when typing


humitos:
I found this while using the search in https://readthedocs-addons.readthedocs.io/

![Peek 2024-08-06 18-01](https://github.com/user-attachments/assets/a5860bdc-380e-42bf-b172-5b55815862c9)

zanderle:
🤦 will get to it asap

PR #361, opened 2024-08-08
Fix weird search behaviour

### LINKED ISSUE #358: Search: it behaves weird 😄
It seems the search is behaving weird when no results are found:

1. it shows `true` (#357)
2. then it shows the recent searches
3. then it shows "Not results found" page

Peek 2024-08-06 18-04

zanderle:
Fixes #358 and #357


src/search.js
```diff
@@ -624,6 +626,7 @@ export class SearchElement extends LitElement {
       let func = () => {
         this.removeAllResults();
       };
+      this.hideRecentSearches = false;
       debounce(func, CLEAR_RESULTS_DELAY)();
```
humitos, line 630: This could confuse users a little bit, since we will show the recent search when we have less than `MIN_CHARACTERS_QUERY`. I think we should show recent search when the input field is completely empty, instead (e.g. `query.length === 0`)

humitos (commented):
It looks good to me. I left some small comments. Let me know.

src/search.js
```diff
@@ -321,6 +321,10 @@ export class SearchElement extends LitElement {
     if (!recentSearches || !recentSearches.length) {
       return html`<p>No recent searches</p>`;
     }
+
+    if (this.hideRecentSearches) {
+      return html``;
```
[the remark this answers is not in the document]
zanderle: Ugh, I've used it many times before, but completely forgot about it here for some reason 😅

src/search.js
zanderle: Hm, I've tried this, but felt the behaviour implemented here felt a bit more natural. Have you tried both options? In any case let me know if you want me to change it, no problem at all.

humitos: I just tested it locally as-is and I think it's fine for now. We can change this behavior in the future if we feel it doesn't work well.

humitos (approved):
Looks good to me 💯

PR #362, opened 2024-08-09
Analytics: remove GA

### LINKED ISSUE #10: Analytics: Read the Docs global
Currently, we are injecting Google Analytics [^1] via Javascript. However, we talked about stop depending on Google Analytics and migrate to Plausible. I'm not sure if Plausible is OK to handle all the traffic we have in all the subdomains `*.readthedocs.io`

[^1]: https://docs.readthedocs.io/en/stable/advertising/advertising-details.html#analytics

humitos:
Closes #10


agjohnson (approved):
This change looks good. Are we not worried about removing this functionality without any sort of deprecation plan (especially with Addons not beta)?

Issue #371, opened 2024-09-09
Integrate with error pages


agjohnson:
The documentation error pages used to direct users to search in our dashboard with an input. I noted here that we should probably rework the search feature to instead use the in-doc search instead:

- https://github.com/readthedocs/readthedocs.org/issues/11486

One other point to integrating with the error pages is that Addons are active on error pages currently -- flyout, version warnings, etc. These feel a little out of place on error pages, and might be conflicting warnings to readers.

agjohnson:
We should disable the flyout and notifications on error pages probably, otherwise they both show by default:

![image](https://github.com/user-attachments/assets/9521ded3-c884-4a4e-85f6-551ee18a35d8)

We want analytics addon to always run though, and might use search in these UIs eventually. So we shouldn't disable all addons on error pages.

humitos:
> I noted here that we should probably rework the search feature to instead use the in-doc search instead:

👍🏼 

> We should disable the flyout and notifications on error pages probably, otherwise they both show by default
> We want analytics addon to always run though, and might use search in these UIs eventually. So we shouldn't disable all addons on error pages.


URL where you can notice this behavior: https://dev.readthedocs.io/en/latest/notfound.html

I've tried this some time ago and I found that's not possible to know what's the status code for the original documentation URL from JavaScript. I will add the status code using CF Worker and reading it from JavaScript following the pattern we are using to read other similar data we don't have access from JavaScript.

humitos:
I opened these PRs to solve this issue:
* https://github.com/readthedocs/addons/issues/371
* https://github.com/readthedocs/common/pull/222

PR #373, opened 2024-09-18
Allow disabling addons on 404 pages

humitos:
It reads a META tag `readthedocs-http-status` to get the HTTP status code for the original request to the documentation URL.

It uses this HTTP status code to decide if each of the addons should be enabled or not.

* Closes https://github.com/readthedocs/addons/issues/371
* Requires https://github.com/readthedocs/common/pull/222


agjohnson (approved):
Looks great!

Perhaps eventually we'll want to pass around a `request` object instead of just the HTTP status, but let's just keep this in mind for now.

humitos:
Hmm. Interesting... How could we inject the whole response/request object from CF? Could we pickle it somehow? If you have an idea, please open an issue explaining this so we can discuss/track it there

Having these objects around would be useful for other stuff for sure.

agjohnson:
Not sure actually. I suppose if we find other users looking for other response attributes we can figure that out. I might just be thinking of a simple object we maintain to consolidate arguments we're passing around now, but I also don't have any other strong recommendations for other data we might need to pass around.

humitos:
> I might just be thinking of a simple object we maintain to consolidate arguments we're passing around now

Good point. I  opened https://github.com/readthedocs/addons/issues/376 to discuss/track this work.

PR #374, opened 2024-09-18
Notification: use `toast` notification by default

humitos:
Use `toast` notification by default. It will be at the top right and smaller than the current approach.

Screenshot_2024-09-18_19-40-55


ericholscher (approved):
Simple enough!

agjohnson (approved):

PR #375, opened 2024-09-18
Notification: auto-dismiss while visible

### LINKED ISSUE #366: Notification: auto-dismiss
Ideally, it should have the following features:

* Using the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) allows us to dismiss the notification only if the tab is in view.
* Pausing the dismiss timer on mouse hover allows for reading of the text and clicking links.
* Visual feedback on the timer helps readability too

Ideas grabbed from https://github.com/readthedocs/addons/issues/46#issuecomment-1706968350

humitos:
It detects if the tab is opened and trigger a timeout for 5 seconds. After that time, the notification is hidden.

Eventually, we could use some nice effect, like fade out and stop the timer while the pointer is over the notification -- but that't outside the scope of this initial work.

Peek 2024-09-18 20-17

Closes #366


src/notification.js
```diff
@@ -58,6 +61,39 @@ export class NotificationElement extends LitElement {
...
+      clearTimeout(this.timerID);
+      this.timerID = null;
+    } else {
+      // Page became visible.
+      // Trigger a timeout to hide the notification.
+      this.timerID = setTimeout(() => {
+        this.autoDismissed = true;
+        this.requestUpdate();
+      }, 5000);
+    }
```
agjohnson, line 77: I might break these two conditions out to individual functions. To support not hiding the notification on hover, we'd reuse this same logic: on `mouseover` & `focusin` clear the timer, and on `mouseout`/`mouseleave` and `focusout` add a new timer.

I probably wouldn't wait until later to support mouse/focus support for pausing the timer, it's a small addition and slightly better UX

agjohnson (approved):
This looks good! I think we should aim to support hover/focus with this change, which shouldn't be much more more work. We'd need to add the events to the templates and call this same logic from those events.

humitos:
I'll try something tomorrow. I'd like to not reset the time, but pause it. I may need to save the countdown time for that... Not sure how yet

agjohnson:
Either works, but resetting seems fine and doesn't require more state tracking. To make a notification that can be paused you could turn the timer into a 1s timer and accumulate a count of seconds unpaused instead.

src/notification.js
humitos: I added support for this in 22734f8

src/notification.js
```diff
@@ -58,6 +64,66 @@ export class NotificationElement extends LitElement {
...
+  _handleMouseEnter = (e) => {
+    // Stop the timer when notification is hovered (mouseenter event)
+    clearTimeout(this.timerID);
+    this.timerID = null;
+  };
+
+  _handleMouseLeave = (e) => {
+    // Start the timer when the notification is hovered away (mouseleave)
+    this.triggerAutoDismissTimer();
+  };
```
agjohnson, line 91: @humitos Note, this doesn't handles the focusin/focusout events I mentioned. These are required for keyboard and screen reader support, which won't fire mouse events.

PR #379, opened 2024-09-19
New Tooltips implementation

### LINKED ISSUE #20: `hoverxref` new name
> I think it could be a good opportunity to re-brand this feature and find a good name for it --not only because the `hoverxref` name is terrible, but also because it's not tied to "xref" anymore, since that's a Sphinx concept

_(from Slack)_

We should find some good ideas for this name and more forward with one. I'm using `tooltips.js` for now.

### LINKED ISSUE #239: Tooltips: `sphinx-hoverxref` popup implementation
I couldn't find an issue where we talked about this specifically, but it has come up in a few calls.

The problem we have currently with hoverxref is that the popup library implementation needs to be replaced as we port it to an addon. I believe the issue here was that the library was a jQuery library? @humitos is this correct?

Options we should weigh:

- Load jQuery and library conditionally, via dynamic webpack import. We still use jQuery in this case, but only when the documentation project uses hoverxref
- We replace the library with another implementation
- We rely on upcoming HTML features that aren't yet fully standard [^1]

[^1]: The [HTML `popover` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) is upcoming, and is currently part of [Interop 2024](https://webkit.org/blog/14955/the-web-just-gets-better-with-interop/#popover). Firefox is the outlier here currently, though there is an implementation in nightly. A hybrid approach would be to do a polyfill or something similar for unsupported browsers.

zanderle:
Closes #239 
Closes #20


src/tooltips.js
```diff
@@ -1,72 +1,183 @@
...
+    cancelShowDelay();
+    const tooltip = getRelatedTooltip();
+    if (hideTooltipTimeoutId === null) {
+      hideTooltipTimeoutId = setTimeout(() => { hideTooltip(); }, HIDE_TOOLTIP_DELAY);
+    }
+  }
+
+  function showTooltip() {
+    // First hide any other tooltips
+    const existingTooltips = document.querySelectorAll("div[data-tooltip-href]");
```
humitos, line 66: It would be good to put this `data-tooltip-href` into a constant since we set the attribute below and we query it here.

src/tooltips.js
```diff
@@ -1,72 +1,183 @@
...
+    }
+  }
+
+  function hideTooltip() {
+    const tooltip = getRelatedTooltip();
+    cancelShowDelay();
+    tooltip.style.display = "";
+  }
+
+  function getRelatedTooltip() {
```
humitos, line 146: I'm not sure to understand why it's called "get related tooltip". I don't get the "related" part of the name. I think I'm missing something here or maybe the name got old from a previous commit? Would it work in the same way as `getTooltip`?

src/tooltips.js
```diff
@@ -1,72 +1,183 @@
...
+      return existingTooltip;
+    } else {
+      const newTooltip = document.createElement("div");
+      newTooltip.insertAdjacentHTML("afterbegin", '<div class="arrow"></div>');
+      newTooltip.insertAdjacentHTML(
+        "afterbegin",
+        '<div class="tooltip-content">Loading...</div>',
+      );
+      newTooltip.setAttribute("data-tooltip-href", anchorElement.href);
+      newTooltip.classList.add("tooltip");
```
humitos, line 167: Is there any reason here why we don't use string literals here?

```javascript
newTooltip = `
  <div class="tooltip" data-tooltip-href="${anchorElement.href}">
    <div class="arrow"></div>
    <div class="tooltip-content">Loading...</div>
  </div>
`;
```

I feel it's a lot easier to read and understand the structure of it.

src/tooltips.js
```diff
@@ -1,72 +1,183 @@
...
+    );
+    if (existingTooltip) {
+      relatedTooltip = existingTooltip;
+      return existingTooltip;
+    } else {
+      const newTooltip = document.createElement("div");
+      newTooltip.insertAdjacentHTML("afterbegin", '<div class="arrow"></div>');
+      newTooltip.insertAdjacentHTML(
+        "afterbegin",
+        '<div class="tooltip-content">Loading...</div>',
```
humitos, line 164: I wasn't able to trigger this "Loading..." in my test. Can you post a screenshot of how it looks?

src/tooltips.js
```diff
@@ -1,72 +1,183 @@
...
+
+    if (newTooltip !== undefined) {
+      const href = anchorElement.href;
+
+      // we set a variable so the data is only loaded once via Ajax, not every time the tooltip opens
+      const hrefAttribute =
+        "http://docs.devthedocs.org/en/stable/tutorial/index.html#modifying-versions";
+      // "http://docs.devthedocs.org/en/stable/tutorial/index.html#upgrading-the-python-version";
+      // "http://docs.devthedocs.org/en/stable/tutorial/index.html";
+      const url = getEmbedURL(hrefAttribute);
```
humitos, line 80: Remove the mocked URLs from here.

src/tooltips.js
```diff
@@ -76,6 +187,9 @@ function getEmbedURL(url) {
     url: url,
   };
 
-  const api_url = "/_/api/v3/embed/?" + new URLSearchParams(params).toString();
+  // const api_url = "https://docs.readthedocs.io/_/api/v3/embed/?doctool=sphinx&doctoolversion=7.3.7&url=https%3A%2F%2Fdocs.readthedocs.io%2Fen%2Fstable%2Fversions.html%23how-we-envision-versions-working";
+  // const api_url = "https://sphinx-hoverxref.readthedocs.io/_/api/v3/embed/?doctool=sphinx&doctoolversion=7.4.7&url=https%3A%2F%2Fsphinx-hoverxref.readthedocs.io%2Fen%2Flatest%2Fhoverxref.html%23hoverxref";
+  const api_url =
+    "https://sphinx-hoverxref.readthedocs.io/_/api/v3/embed/?doctool=sphinx&doctoolversion=7.4.7&url=https%3A%2F%2Fwww.sphinx-doc.org%2Fen%2Fmaster%2Findex.html";
```
humitos, line 193: Hit the correct API URL here.

humitos (commented):
This is working great so far!

Peek 2024-11-05 10-49

There are a few missing pieces we need to do before we are able to merge this work:

- [ ] make tooltips enabled/disabled based on the API response
- [ ] convert the `initializeTooltips` into a web component following the pattern we have for the others
- [x] call the correct API URL instead the mocked one
- [x] get the CSS selector from the API response (eg. `[role=main] a.internal`)
- [ ] write some tests, if possible

I can probably take the work on most of them. @zanderle would you mind helping me with the tests once we have it migrated to a web component?

humitos:
I opened https://github.com/readthedocs/addons/pull/422 with some of the work missing here.

src/tooltips.js
zanderle: Good point

src/tooltips.js
zanderle: You're right. This is a remnant from the first implementation, when it wasn't scoped. The current implementation it would make more sense to drop `related`

src/tooltips.js
zanderle: Hm, good point. I was under the assumption I have to do it element by element when using `insertAdjecentHTML`, but I haven't tested, and now that I think about it, that's probably not the case. I'll try it, and fix it if possible

humitos:
Marking this ready for review so @agjohnson can take a look if he wants.

PR #391, opened 2024-10-02
Search: check if there are filters before render them

humitos:
Before rendering the filters section, we check if there are filters defined in the response API. If there aren't, we don't render this section at all.

Screenshot_2024-10-02_15-42-04


ericholscher (approved):
Looks good. 👍

PR #392, opened 2024-10-03
Flyout: use logo on collapsed and logo+wordmark on expanded

humitos:
We talked about reducing the space in the flyout by removing the "Read the Docs" word mark from the collapsed version of the flyout.

This will give us some room for icon+language implemented in https://github.com/readthedocs/addons/pull/360 and also to display a dropdown (or similar) icon to communicate there are features behind it.

Peek 2024-10-03 12-04


humitos:
I'm merging this PR so we move forward here. @ericholscher feel free to provide any feedback you may have here 👍🏼 . I'm open to perform more adjustments.

humitos:
One thing that I don't like too much is that with the smaller version of the flyout, the ad width and the flyout width is not aligned 😄 

Screenshot_2024-10-15_11-26-41


This is completely subjective and _minor_, but I just wanted to mentioned it 😸

PR #394, opened 2024-10-03
Flyout: show a caret on flyout

### LINKED ISSUE #364: Flyout: make it obvious there are features inside it
> The logo takes up a lot of space and doesn't exactly communicate "click here for hosted features".

_(@agjohnson in https://github.com/readthedocs/addons/pull/360#issuecomment-2286908369)_

> I agree that the logo would benefit from some change. The current state looks more like an advertisement for RTD than a feature of the opened project’s docs; this disturbs users' focus.

_(@orangesunny in https://github.com/readthedocs/addons/pull/360#issuecomment-2288462774)_

----

I'm opening this issue here to talk more about this and find a way to keep the Read the Docs branding on the flyout while communicating there are hosting features _behind a click_ on the it.

There is an issue opened about re-designing the flyout completely as well at #53

### LINKED ISSUE #393: Flyout: communicate there is features in the expanded version
> Also, it's worth noting here that we also dropped the dropdown caret on the right side of the menu. Without this, there isn't a visual cue that the flyout menu expands.

_(from https://github.com/readthedocs/addons/pull/360#issuecomment-2350049920)_

Example of what we had in the old flyout

Screenshot_2024-10-03_12-18-38

humitos:
Small POC to showcase how it could look like #393

Peek 2024-10-03 12-29

Closes #393 
Closes #364


sciencewhiz:
This would also close #364, right?

ericholscher (approved):
This looks like a solid improvement to me. 

We probably want to wait on #400 in order to not overflow the menu even more on long version names, though.

humitos:
Updated with the new code that implements grid flex:

Peek 2025-01-13 15-06

I think this is ready to re-review and merge probably. cc @agjohnson

ericholscher:
Yea, this looks nicer 💯

agjohnson (approved):
Looks good 👍

ericholscher:
If we wanted to get fancy, we should probably have an `up` carrot when closed, and a `down` carrot when open, but I don't feel too strongly.

PR #401, opened 2024-10-10
Make the ad CTA a button

davidfischer:
Note: this only applies to the sidebar or (non fixed) footer ad. My idea here was to test it out and see how this performs before rolling it out on the EA styled ads. Any feedback before rolling it out is also valuable.

## Screenshot
Screenshot from 2024-10-10 15-36-51


humitos:
> My idea here was to test it out and see how this performs before rolling it out on the EA styled ads.

This CSS code won't be required here once it's rolled out in the EA styles?

humitos (approved):
Looks good to me.

Is there a way to test it? I understand the server needs to return an ad with the "Try it out" text now.

Reminder: we need to re-build the assets before merging. I can do that if you don't want to create the environment.

davidfischer:
> This CSS code won't be required here once it's rolled out in the EA styles?

Correct. This is *just* for the RTD styles.

> Is there a way to test it? I understand the server needs to return an ad with the "Try it out" text now.

This should be testable as-is. No server changes are required. This is purely a stylistic change. If there's no CTA on the ad, it won't display a button but we already bold the CTA and this just makes it a button instead.

I've committed the built assets.

humitos:
> This should be testable as-is. No server changes are required.

Ah, cool! I asked because I wasn't receiving any ad with this button.

I checked today, and I got one. It works fine to me.

Screenshot_2024-10-12_14-46-28

humitos:
Ups... I just saw this broken ad:

Screenshot_2024-10-12_14-59-26

ericholscher:
@davidfischer is this from the ad having broken HTML? 🤔

davidfischer:
> @davidfischer is this from the ad having broken HTML? 🤔

I fixed up this individual ad. The advertiser had kind of messed this up and put a CTA with a typo in there (no space between `config.` and `More`)

PR #403, opened 2024-10-11
EthicalAd: show the ad more integrated on Sphinx's furo theme

humitos:
Show the ad in the toc tree container (right bar). I tried adding it to the left bar similar to our own theme, but since they use a narrow (in width) bar, it doesn't look good because it touches the borders.

Screenshot_2024-10-11_19-30-10


davidfischer (approved):
This looks good and should work. It's nice to have something a little more integrated when we can.

humitos:
I may need to check this more. It's suppose that Furo has a explicit placement defined: https://github.com/pradyunsg/furo/blob/fd3e7b94e9959870aa3c8efb221de66be923421d/src/furo/theme/furo/sidebar/ethical-ads.html#L4C7-L4C32

However, the documentation served at https://scanoss.readthedocs.io/en/latest/ does not include this explicit placement. It can be disabled by overwriting `html_sidebars`, but I didn't find a configuration like that in this project. Something else is going on here and I wasn't able to find it.

...


Oh, I see. I think it's because `READTHEDOCS` is not injected in the context anymore, so that block is not included https://github.com/pradyunsg/furo/blob/fd3e7b94e9959870aa3c8efb221de66be923421d/src/furo/theme/furo/sidebar/ethical-ads.html#L2

Ideally, furo should do something like we do in our theme now to define this variable in the context: https://github.com/readthedocs/sphinx_rtd_theme/blob/7b894f5e7a5053dd35de406bc779020c229f20a9/sphinx_rtd_theme/__init__.py#L55-L65

In the meanwhile, I think we can add the `id=furo-sidebar-ad-placement` ourselves in the div we inject for now until they solve this issue.

ericholscher (commented):
This is a reasonable stopgap :+1:

PR #407, opened 2024-10-14
Notification: don't show it on `Project.default_version`

### LINKED ISSUE #406: Notification: "default version" shouldn't show "This may be an outdated version"
> If the "default version" could be thrown in the mix as a version that's also not outdated we'd be happy. The assumption here of course being that if a project sets a version as the default it's not likely to be outdated. 

_(posted by @ kbevers in https://github.com/readthedocs/addons/issues/402#issuecomment-2409993674)_

----

I'm opening this issue to research about this a little more.

@kbevers do you have an example where you getting the unexpected behavior? That would be pretty useful to me to understand what's going on here.

humitos:
Don't show the notification "This may be an old version of this documentation" if the reader is currently reading the `Project.default_version`.

I'm on the fence about doing this part of the default logic or a config on its own. I'll probably change it to a config after thinking a little more, but pushing these changes for now.

Closes #406


ericholscher:
I really don't like having so many configuration options. I already feel like the Addons config is overwhelming, and "just add another setting" feels like the wrong solution.

I think users almost certainly don't want a warning on their default version, at least in the current UX of the warning, so I'm :+1: on this being part of the default logic.

humitos:
Cool! I updated the PR and I'm marking it ready to review.

PR #414, opened 2024-10-29
Notifications: allow disable `latest` and/or non-`stable`

### LINKED ISSUE #133: Notification: allow to disable `latest` and/or non-`stable` warning
The current pretty simple logic we have is good and it may work for most of the cases. However, it's a little verbose since _it will show a notification in all the versions that are not `stable`_ --which could be _a lot_.

There are users that won't like this pattern and may not want to follow "our proposal to use the platform" here (most users should always read the `stable` version of the project's documentation) or may require an un-supported way to decide "stable" version.

These users should be able to decide whether or not show a warning notification on `latest` and/or on non-`stable` versions independently:

* Enable the addons (disable it completely)
* Show a notification on non-`stable` versions
* Show a notification on `latest` version

Related

* #125
* https://github.com/readthedocs/ext-theme/issues/212
* https://github.com/readthedocs/ext-theme/issues/211
* https://github.com/cilium/cilium/issues/29969
* https://github.com/readthedocs/addons/issues/201

humitos:
Closes #133
Requires https://github.com/readthedocs/readthedocs.org/pull/11718


humitos:
Backend and frontend PRs are already approved. This is only one missing for review to be able to merge them all.

agjohnson (approved):

Issue #413, opened 2024-10-29
Use namespaced class names for CSS classes on the host element


laymonage:
Hey all.

I've spent a few hours debugging why we're not seeing the notifications for old versions and PR builds with the new Notification add-on. It turns out that the element _is_ there, but Bootstrap (which our theme uses) hides it because it has the `toast` class:

<img width="1794" alt="hidden notification element due to opacity:0 set by bootstrap" src="https://github.com/user-attachments/assets/a3038c8c-e77e-434a-b8ef-eb2df2af876f">

See the `opacity: 0;` at the bottom right of the screenshot. That comes from Bootstrap, and it requires toast to be initialised via JS (hence the initial state has `opacity: 0;`).

Debugging this was quite hard because the notification immediately disappears after 5s. I had to overwrite the JS and increase the timeout to investigate this.

Is there a reason why the custom element needs to rely on CSS classes in the host element? Could you perhaps use something else, e.g. data attributes?

Or, if you want to use CSS classes, I think they should be namespaced, e.g. prefixed with `rtd-` so that it doesn't conflict with the theme's CSS (and can still be overridden if that's the desired goal).

Thanks!

P.s. love the addons – DocDiff in particular is really awesome.

humitos:
Thanks for opening this issue. I understand this should not happen since the notification is a Web Component with its shadow DOM is closed. I will need to take a deeper look to understand what's going on.

> Is there a reason why the custom element needs to rely on CSS classes in the host element? Could you perhaps use something else, e.g. data attributes?

@agjohnson shouldn't the web component ignore this external modification? Is it maybe because we left the shadow DOM open somehow?

agjohnson:
I can't see the notification on this example, so can't tell you what is happening yet.

But yes, the elements we inject are inside a shadow DOM and should not inherit CSS styling from the document root DOM. We do allow variable customization of some of the CSS properties, all of those variables are prefixed with `--readthedocs-...`.

My guess is there is something else happening here.

laymonage:
@agjohnson Demo'd the issue in #416.

https://readthedocs-addons--416.org.readthedocs.build/


https://github.com/user-attachments/assets/60a80c53-673f-4ece-a213-b9aaa9c5d1c3

laymonage:
[...]

PR #416, opened 2024-10-30
Demonstrate CSS from docs leaking into notification component

laymonage:
Hopefully this should demo the issue in #413.

https://github.com/user-attachments/assets/60a80c53-673f-4ece-a213-b9aaa9c5d1c3

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,23 @@
+.toast {
+    animation-name: spin;
+    animation-duration: 500ms;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+    opacity: 0.5;
+    font-weight: bold;
+    animation-name: spin;
+    position: fixed;
+    display: block;
+    top: 0;
+    right: 0;
+}
+
+@keyframes spin {
+    from {
+        transform: rotate(0deg);
+    }
+
+    to {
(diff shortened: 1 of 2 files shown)


docs/_static/custom.css
```diff
@@ -0,0 +1,23 @@
+.toast {
```
laymonage, line 1: In our case, the styling for `.toast` comes from Bootstrap which is used by the theme. The styling is not meant to be applied to RTD's notification web component, but it gets applied anyway because the injector applies the class on the host `<readthedocs-notification>` component, which lives on the root `<html>` document instead of the shadow DOM.

laymonage: I don't see the reason why it needs to apply these classes though?

https://github.com/readthedocs/addons/blob/6c86cb04642b797e992235f8ea7e103de3e757ff/src/notification.js#L170-L174

Perhaps it is intentional that you allow the component to be styled differently by changing the CSS classes that are applied on the host component. For example, allowing different variants simply by toggling the class names on the `<readthedocs-notification>`. If so, then you should use namespaced class names, because the host element lives in the root `<html>` element. Or, use something else with even less chance of clashing e.g. data attributes. 

This applies to all the class names in `:host(...)` in https://github.com/readthedocs/addons/blob/main/src/notification.css.

PR #422, opened 2024-11-05
Implement tooltip as an addon

humitos:
I did some of the work mentioned in https://github.com/readthedocs/addons/pull/379#pullrequestreview-2415147994

However, since we were using scoped functions for each of the elements, we will need to rewrite that code a little now that we are using a web component. Another idea would be to define `readthedocs-tooltip` as _one_ tooltip and add one per link to have it self-contained.

@agjohnson @zanderle do you have any opinion here?


zanderle:
> I did some of the work mentioned in [#379 (review)](https://github.com/readthedocs/addons/pull/379#pullrequestreview-2415147994)
> 
> However, since we were using scoped functions for each of the elements, we will need to rewrite that code a little now that we are using a web component. Another idea would be to define `readthedocs-tooltip` as _one_ tooltip and add one per link to have it self-contained.
> 
> @agjohnson @zanderle do you have any opinion here?

My understanding was that the requirement is to have the links be defined as normal `a` elements, and have the javascript make them into tooltips. With this in mind, I don't see why we would need a web component.

However if we did `readthedocs-tooltip`, then it would make sense to do it as a web component.

To me it boils down to: how do we want this to be used in the end? As normal `a` elements or as `readthedocs-tooltip`? And that will determine the implementation we choose.

zanderle:
> > My understanding was that the requirement is to have the links be defined as normal `a` elements, and have the javascript make them into tooltips. With this in mind, I don't see why we would need a web component.
> 
> Yeah, this is what we want.
> 
> I got confused while migrating into the structure we already have. I don't think we _need_ a `readthedocs-tooltip` web component; but we need somehow to follow the pattern we already have for the other ones. This pattern will be similar to `EthicalAdsAddon` that extends from `AddonBase` but doesn't declare a web component at all.
> 
> I will try copy that pattern here as well.

I see what you mean - we still need a way to load the config, and do stuff based on that, right?

That's a good point, but yeah, I'm not sure a web component is needed for that.

 I can handle it if you want?

humitos:
@zanderle I just pushed a new commit to this branch that implements it as an addon. Please, review it, and if you think it's fine, we can merge it into your PR and then merge yours as well 👍🏼 

I just tested it locally and it works great!

It won't be enabled by default because the API won't return the `addons.tooltips` key yet. We will control the tooltip addon from there by opening a new PR in the server.

zanderle (commented):
Looks good 👍 Nice to see all those improvements so quickly :) 

One thing I would probably change is to keep `setupTooltip` as a separate scoped function, and then just call it from within `TooltipsAddon` (vs implementing it as a method, as you did here). Similar to how we do other addons - in other addons we first define the `LitElement`, and then call/initialize it from within the specific Addon. I'd do a similar thing here. Except instead of `LitElement`, we'd have `setupTooltip`. And we'd probably want to make `setupTooltip` accept `config` in addition to `el`.

humitos:
Do you mean putting `setupTooltip` _outside_ the `TooltipsAddon` definition? Like, defined "globally" in the same file?

zanderle:
> Do you mean putting `setupTooltip` _outside_ the `TooltipsAddon` definition? Like, defined "globally" in the same file?

Exactly

zanderle:
I just saw the additional question :)

> If that's the case, what are the pros/cons of each approach?

It's not a big difference, but I like to have it separate to be more consistent with the other addons. And if the `setupTooltip` is a method in the `TooltipAddon` then `this` applies to the `TooltipAddon` - which could be a convenient thing, but it also means some shared state between the tooltips, which might not be what we want. My personal preference/opinion is that it would be a bit safer to keep it separate. However I don't think there's right or wrong approach here.
It would also be easier to use outside the addon, if that was ever needed.

Issue #409, opened 2024-10-17
File tree diff client


stsewd:
With https://github.com/readthedocs/readthedocs.org/pull/11646/, we are now exposing the file tree diff information to addons, but we still need to have a way to show the results to users.

What I'm thinking here is to have a small list of the changed files (the first 2 or 3 files of each category), and then have a "show all files" button that will open a popup with the list of the files, this list will have some options to filter the results (modified, added, deleted) and a search bar.

We can also expose a button that suggest how to create a redirect for files that were deleted.

ericholscher:
I think to start, even just outputting the data to the console, or as a `<details>` tag in the PR build warning seem like a good initial implantation. Just something where we can start surfacing the data to start playing with in dev.

TheTripleV:
This is nice to have. We currently have a sphinx extension that does this for us with git. https://github.com/wpilibsuite/sphinxext-delta

It injects a new main toctree section that lists all the pages that have changed.

https://frc-docs--2680.org.readthedocs.build/en/2680/index.html

is an example of a PR build with the "pr changed files" toctree added.
We're currently fixing failures in this extension because RTD doesn't inject all the html context anymore.

humitos:
I checked the [API response](https://docs--11716.org.readthedocs.build/_/addons/?client-version=0.19.0&api-version=1&url=https%3A%2F%2Fdocs--11716.org.readthedocs.build%2Fen%2F11716%2Fintro%2Fadd-project.html&project-slug=docs&version-slug=11716) returned at https://docs--11716.org.readthedocs.build/en/11716/

```
        "filetreediff": {
            "diff": {
                "added": [
                    {
                        "file": "guides/management/2fa.html"
                    }
                ],
```

It seems the `file` attribute is not a valid URL, but the path of the file on disk that changed. We will need to resolve all those paths into valid URLs in the backend and expose them to the frontend to generate the links from JavaScript. We can probably add a `url` attribute in the API response as well.

PR #424, opened 2024-11-07
File Tree Diff: initial implementation

humitos:
Added a new addon to show all the filenames and URLs that were added, deleted or modified in the current PR compared to the LATEST version.

Screenshot_2024-11-07_12-44-02

I'm rendering this data in a new notification decoupled from the existing one on purpose. I don't want to entangle the code of these two different addons for now, in particular while we are under development/testing. I also see that in the future they may have different UI elements, or the data will be integrated in a more bigger UI element (eg. toolbar) or similar.

We can discuss more how this addon will grow and where is a good place to add this data and make that work in a following PR. For now, I want something that we can start testing internally in our own projects.

### Examples

Screenshot_2024-11-07_14-15-37

Screenshot_2024-11-07_14-35-41

Screenshot_2024-11-07_14-51-06

Closes https://github.com/readthedocs/addons/issues/409
Requires https://github.com/readthedocs/readthedocs.org/pull/11749


humitos:
### Usage example (together with docdiff)

Peek 2024-11-07 15-13

ericholscher (approved):
This looks like a great start. I'm :+1: on shipping it overall. I'm not well versed on the JS style other than my silly comment below :)

src/filetreediff.css
```diff
@@ -0,0 +1,86 @@
+:host > div {
+  margin: 1rem 0rem;
+  padding-top: 1rem;
+  padding-bottom: 1rem;
+  overflow: auto;
+  border-radius: 0.5rem;
+  font-family: "Lato";
```
agjohnson, line 7: You should use a more permissive font family, as Lato isn't hosted by the Addons and it isn't commonly installed for end users. This is an okay default:

https://github.com/readthedocs/addons/blob/0f025e3bf604f96506aebf2af120b4c20d02fde8/src/notification.css#L14C1-L21C5

agjohnson (approved):
This all looks great! I think this is good to start testing.

We have had some related conversations recently that would change this work -- using attributes over classes, naming of this feature -- but don't think these need to hold up our internal testing.

PR #429, opened 2024-11-11
Remove check for old integration

humitos:
We are not injecting `readthedocs-doc-embed.js` anymore and it's being removed by the CF worker anyways. This is dead code now.


agjohnson (approved):
Code looks good. Just to clarify, this code isn't used anymore as even if projects still have this old include, our addons injection is removing the script anyways. Is that the assumption here? This seems good if so.

humitos:
> this code isn't used anymore as even if projects still have this old include, our addons injection is removing the script anyways. Is that the assumption here?

Yes.

PR #433, opened 2024-11-14
Link previews: updates after meeting

humitos:
1. Apply link previews only on links from the same domain as current
2. Auto detect the doctool used
3. Create a specific selector based on doctool detected
4. Allow overriding the CSS selector if it's sent in the API

Related https://github.com/readthedocs/readthedocs.org/pull/11767


src/linkpreviews.js
```diff
@@ -224,13 +224,20 @@ export class LinkPreviewsAddon extends AddonBase {
...
-      "[role=main] a.internal",
-    );
+    const rootSelector = this.config.addons.linkpreviews.root_selector;
+
+    let selector;
+    if (doctoolName === "sphinx") {
+      selector = `${rootSelector} a.internal`;
+    } else {
+      selector = `${rootSelector} a`;
+    }
```
agjohnson, line 234: Also, if we need it, `querySelector*` allow for compound selectors too:

```js
${rootSelector} a.internal, ${rootSelector} a
```

This doesn't work for Sphinx, but I could see the generic selector growing like this.

src/linkpreviews.js
```diff
@@ -224,13 +224,20 @@ export class LinkPreviewsAddon extends AddonBase {
...
 
     const elements = document.querySelectorAll(selector);
-    elements.forEach((el) => setupTooltip(el, doctoolname, doctoolversion));
+    for (const element of elements) {
+      if (new URL(element.href).hostname === window.location.hostname) {
```
agjohnson, line 238: The call to `new URL()` could use some error handling. `URL()` requires a full URL, and though browsers _should_ normalize `<a href="/en/latest">` to a full URL, there are likely some edge cases to cover here that would cause an error at `new URL()`.

agjohnson (approved):

src/linkpreviews.js
```diff
@@ -210,12 +210,12 @@ export class LinkPreviewsAddon extends AddonBase {
...
+    const doctoolName = objectPath.get(
       this.config,
       "addons.linkpreviews.doctool.name",
       "unknown",
     );
-    const doctoolversion = objectPath.get(
+    const doctoolVersion = objectPath.get(
       this.config,
       "addons.linkpreviews.doctool.version",
       "unknown",
```
humitos, line 221: Taking into account what we talked in https://github.com/readthedocs/readthedocs.org/pull/11767. What should we do with these fields? cc @agjohnson @ericholscher 

Note that we have specific code for Sphinx in our EmbedAPI backend: https://github.com/readthedocs/readthedocs.org/blob/44587ab35067dd36cf4fcd94726acf70a47f2163/readthedocs/embed/v3/views.py#L166-L270

src/linkpreviews.js
humitos: I updated the code for this. I'm not sure _what_ to do with the error, I'm just skipping that element for now. Let me know if you have a better idea.

src/linkpreviews.js
humitos: Also, we will need to do something with the following code as well: https://github.com/readthedocs/addons/pull/433/files#r1842484521

ericholscher: > Taking into account what we talked in [readthedocs/readthedocs.org#11767](https://github.com/readthedocs/readthedocs.org/pull/11767). What should we do with these fields? cc @agjohnson @ericholscher
> 
> Note that we have specific code for Sphinx in our EmbedAPI backend: https://github.com/readthedocs/readthedocs.org/blob/44587ab35067dd36cf4fcd94726acf70a47f2163/readthedocs/embed/v3/views.py#L166-L270

Seems like we can keep this logic? I thought we discussed having the ability to set it, just to not make it super visible in the UI? 


> Also, we will need to do something with the following code as well: https://github.com/readthedocs/addons/pull/433/files#r1842484521

Not sure what this is referring to?

humitos: > Seems like we can keep this logic? I thought we discussed having the ability to set it, just to not make it super visible in the UI?

In https://github.com/readthedocs/readthedocs.org/pull/11767, we are talking about removing the ability for the user to set the "documentation tool". That's my point. If we are going to keep it, it should be "global", if we are going to remove it, we need to think what to do in this case.



> Not sure what this is referring to?

This links to a comment on this PR that defines the CSS selector based on the documentation tool.

ericholscher: Well, if we remove the tool from the backend, presumably we'd just stop sending it?

src/linkpreviews.js
```diff
@@ -224,13 +224,27 @@ export class LinkPreviewsAddon extends AddonBase {
...
+    } else {
+      selector = `${rootSelector} a`;
+    }
 
     const elements = document.querySelectorAll(selector);
-    elements.forEach((el) => setupTooltip(el, doctoolname, doctoolversion));
+    for (const element of elements) {
+      try {
+        let elementHostname = new URL(element.href).hostname;
+        if (elementHostname === window.location.hostname) {
```
ericholscher, line 240: Seems this part could be removed from the other discussions, and merged as is? I think we should probably revert those changes, and at least get this change in, and we can discuss the larger changes in another PR?

src/linkpreviews.js
humitos: I'm filtering out those links that point to the _same page_ the user is currently reading.

src/linkpreviews.js
```diff
@@ -210,27 +210,52 @@ export class LinkPreviewsAddon extends AddonBase {
...
-    );
-    const doctoolversion = objectPath.get(
-      this.config,
-      "addons.linkpreviews.doctool.version",
-      "unknown",
+    // Autodetect if the page is built with Sphinx and send the `doctool=` attribute in that case.
+    let doctoolName = "unknown";
+    const sphinxHtmlElement = document.querySelector(
+      "html[data-content_root], html.writer-html5",
     );
```
humitos, line 217: This auto-detection could be improved eventually, but it works for now.

src/docdiff.js
```diff
@@ -88,8 +88,7 @@ export class DocDiffElement extends LitElement {
       return;
     }
     this.config = config;
-    this.rootSelector =
-      this.config.addons.doc_diff.root_selector || "[role=main]";
+    this.rootSelector = this.config.addons.options_root_selector;
```
agjohnson, line 91: And same as the link preview root selector, using the list of selectors as a complex selector is also not going to work as we're intending for diffing.

I would have to look up how `querySelector` decides which element to return, but I can say it's probably not what we want:

image

humitos:
I updated this PR by reusing all the auto-detect logic we were using for EthicalAds. With that logic, we are now using specific CSS selector based on the documentation tool detected -- if we don't detect any doctool, we iterate over generic CSS selectors and stop after finding the first one.

The user can override the CSS root selector to avoid using the auto-detection logic by specifying a value in from the Web UI.

src/utils.js
```diff
@@ -257,3 +257,245 @@ export function getLinkWithFilename(url) {
...
+    this.DEFAULT_ROOT_SELECTOR = {
+      sphinx: "[role=main]",
+      "mkdocs-material": "main > div > div.md-content",
+      docsify: "article#main",
+      asciidoctor: "div#content",
+      pelican: "article",
+      docusaurus: "article",
+      antora: "article",
+      jekyll: "article",
+      fallback: ["main", "div.body", "div.document", "body"],
```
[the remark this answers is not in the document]
agjohnson: Yeah the deeper we go here, the easier it's seeming to just detect where the content actually is in the DOM. I looked through the readability API and it seemed to do a little bit on top of just detection, but there is probably a way to reuse this or to maybe help shape the API for our use case.

src/utils.js
```diff
@@ -257,3 +267,265 @@ export function getLinkWithFilename(url) {
...
+      ANTORA: "article",
+      JEKYLL: "article",
+      FALLBACK_DOCTOOL: ["main", "div.body", "div.document", "body"],
+    };
+
+    this.DEFAULT_LINK_SELECTOR = {
+      sphinx: "a.internal",
+      fallback: ["p a"],
+    };
+  }
```
agjohnson, line 298: These seem like class static variables, not instance variables:

```js
export class DocumentationTool {
    static DEFAULT_ROOT_SELECTOR = {
      SPHINX: "[role=main]",
      MKDOCS_MATERIAL: "main > div > div.md-content",
      DOCSIFY: "article#main",
      ASCIIDOCTOR: "div#content",
      PELICAN: "article",
      DOCUSAURUS: "article",
      ANTORA: "article",
      JEKYLL: "article",
      FALLBACK_DOCTOOL: ["main", "div.body", "div.document", "body"],
    };

    static DEFAULT_LINK_SELECTOR = {
      sphinx: "a.internal",
      fallback: ["p a"],
    };

    something() {
      console.log(this.constructor.DEFAULT_ROOT_SELECTOR);
    }
  }
```

src/utils.js
```diff
@@ -257,3 +267,265 @@ export function getLinkWithFilename(url) {
...
+      DOCUSAURUS: "article",
+      ANTORA: "article",
+      JEKYLL: "article",
+      FALLBACK_DOCTOOL: ["main", "div.body", "div.document", "body"],
+    };
+
+    this.DEFAULT_LINK_SELECTOR = {
+      sphinx: "a.internal",
+      fallback: ["p a"],
+    };
```
agjohnson, line 297: Unless there is a reason to not match the pattern above, this could use the constants too:

```js
static DEFAULT_LINK_SELECTOR = {
  SPHINX: "a.internal",
  FALLBACK_DOCTOOL: ["p a"],
}
```

src/utils.js
```diff
@@ -257,3 +267,265 @@ export function getLinkWithFilename(url) {
...
+    ) {
+      return `${this.getRootSelector()} ${
+        this.DEFAULT_LINK_SELECTOR[this.documentationTool]
+      }`;
+    }
+
+    // Fallback to our list of generic selectors and stop in the first we found.
+    let element;
+    let fullSelector;
+    for (const selector of this.DEFAULT_LINK_SELECTOR.fallback) {
```
agjohnson, line 320: Assuming class static variable and we want to reuse the constants we started using:

```suggestion
    for (const selector of this.constructor.DEFAULT_LINK_SELECTOR[FALLBACK_DOCTOOL]) {
```

src/utils.js
```diff
@@ -257,3 +267,265 @@ export function getLinkWithFilename(url) {
...
+    if (
+      this.documentationTool &&
+      objectPath.get(this.DEFAULT_ROOT_SELECTOR, this.documentationTool, null)
+    ) {
+      return this.DEFAULT_ROOT_SELECTOR[this.documentationTool];
+    }
+
+    // Fallback to our list of generic selectors and stop in the first we found.
+    let element;
+    for (const selector of this.DEFAULT_ROOT_SELECTOR.fallback) {
```
agjohnson, line 349: Same as above here, static variable access and use the constant for fallback name.

src/utils.js
```diff
@@ -257,3 +267,265 @@ export function getLinkWithFilename(url) {
...
+      'link[href^="_static/celery.css"]',
+      'link[href^="_static/babel.css"]',
+      'link[href^="_static/platter.css"]',
+      'link[href^="_static/werkzeug.css"]',
+    ];
+    for (const selector of selectors) {
+      if (document.querySelectorAll(selector).length === 1) {
+        return true;
+      }
+    }
```
agjohnson, line 474: This looks like a place where we don't need to iterate. The loop is just looking for the first hit in the list of selectors, which querySelectorAll will do for us.

```suggestion
    if (document.querySelectorAll(selectors.join(", ")).length) {
      return true;
    }
```

agjohnson (approved):
This approach looks good! I've noted just some code level changes, good to merge after those though.

src/utils.js
humitos: He, I did this originally, but I wasn't able to figure it out how to use them. I was missing the `.constructor` part. I didn't want to deal with that and used instance variable to be able to move forward 😄

PR #438, opened 2024-11-18
Small DX improvements

ericholscher:
This is mostly quality of life stuff I hit when doing dev.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,6 +38,11 @@ export function setup() {
           if (addon.requiresUrlParam()) {
             sendUrlParam = true;
             break;
+          } else {
+            // Log that the addon isn't going to load because of URL
+            console.debug(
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -38,6 +38,11 @@ export function setup() {
           if (addon.requiresUrlParam()) {
             sendUrlParam = true;
             break;
+          } else {
+            // Log that the addon isn't going to load because of URL
+            console.debug(
+              `Addon ${addon.addonName} is not going to load because it is not enabled for this URL.`,
```
humitos, line 44: I think the logic is not correct here. We need to check if the `url=` param is in the requested URL or not before showing this message.

humitos (commented):
Looks good, but the logic for the URL param needs to be adjusted.

ericholscher:
Gonna close this, and fix it in #439 which was branched off here.

PR #440, opened 2024-11-19
Don't wrap the flyout header.

ericholscher:
This makes it so the version doesn't overflow as badly:

## Before
Screenshot 2024-11-19 at 2 30 24 PM

## After 

Screenshot 2024-11-19 at 2 30 12 PM

### DIFF
--- a/src/flyout.css
+++ b/src/flyout.css
@@ -55,7 +55,7 @@
 
 header {
   display: flex;
-  flex-flow: row wrap;
+  flex-flow: row nowrap;
   justify-content: space-between;
   align-items: center;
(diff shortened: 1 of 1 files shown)


agjohnson:
Unfortunately, both options have issues. I don't think we should shrink the flex grid down without addressing content widths of the labels

While at least it doesn't appear under the logo, the long strings now wrap oddly and can look worse with icons:

image

We should truncate the text to fix this: https://css-tricks.com/flexbox-truncated-text/

humitos:
Closing as this was done in #486

Issue #103, opened 2023-08-31
Docdiff: cache the page downloaded


humitos:
If you press multiple times `d` on the page, each time it will download the "source page" to compare with the current one. It would be good to cache the downloaded DOM.

PR #441, opened 2024-11-19
DocDiff: Cache remote content

ericholscher:
Fixes https://github.com/readthedocs/addons/issues/103


src/docdiff.js
```diff
@@ -120,6 +121,11 @@ export class DocDiffElement extends LitElement {
   }
 
   compare() {
+    if (this.cachedRemoteContent) {
+      this.performDiff(this.cachedRemoteContent);
```
agjohnson, line 125: `performDiff` can throw an exception and this is now outside the `fetch` promise that would normally handle exceptions that bubble up.

However, it's also not great to mix async/promises and sync code, eventually this causes problems. I'd add this to the promise chain as, which if you haven't done async JS does require some extra steps.

Particular, treating `this.cachedRemoteContent` as a promise by splitting this up conditionally:

```js
let promiseData;
if (this.cachedRemoteContent !== null) {
  promiseData = Promise.resolve(this.cachedRemoteContent);
} else {
  promiseData = fetch(...)
    .then((response) => {
      ...
      return response.text();
    })
}

promiseData.then((text) => {
  this.cachedRemoteContent = text;
  this.performDiff(text);
}).catch((error) => { ... });
```

ericholscher: 🤯

src/events.js
```diff
@@ -2,9 +2,7 @@ import { ADDONS_API_VERSION, getMetadataValue } from "./utils";
 
 export const EVENT_READTHEDOCS_SEARCH_SHOW = "readthedocs-search-show";
 export const EVENT_READTHEDOCS_SEARCH_HIDE = "readthedocs-search-hide";
-export const EVENT_READTHEDOCS_DOCDIFF_ADDED_REMOVED_SHOW =
-  "readthedocs-docdiff-added-removed-show";
-export const EVENT_READTHEDOCS_DOCDIFF_HIDE = "readthedocs-docdiff-hide";
+export const EVENT_READTHEDOCS_DOCDIFF_TOGGLE = "readthedocs-docdiff-toggle";
```
agjohnson, line 5: Why were the events changed from show/hide to toggle?

It seems like we still want show/hide here, as in an evented interface there is no feedback whether the flyout/diff/search/etc is in a showing or hidden state.

ericholscher: I noted this above, but the state was stored in 2 places, and broke because we now have ways of activating docdiff without the hotkeys.

ericholscher: But the hotkeys doesn't keep the state to know which event to fire, so you have to just fire the toggle event, otherwise it has weird behavior like doing nothing the first time you fire a "open" event when the keyboard arg has already opened it.

agjohnson: What I described above updates `this.docDiffShow` to keep track of that state inside the hotkeys web component. The hotkey element can still use `this.docDiffShow` to determine which event it's firing then.

The URL param automatically showing the diff should also fire this event.

ericholscher: > The URL param automatically showing the diff should also fire this event.

This would fix it.

Though it still seems very odd to me that the hotkeys would keep track of any state. It seems like the addon that is being effected should track state, and the hotkeys are just triggering hotkey events?

humitos: I think we want to have and use the 3 events: show, hide and toggle. `readthedocs-diff=true` should trigger `SHOW` event, because it want to enable docdiff _always_, no matter what's the state it was previously.

Besides, HotKeys addon should trigger `TOGGLE` because it depends on the previous state to work, and it shouldn't keep track of any state of any addon.

I imagine `HIDE` will be eventually used when for example we want to "disable/hide/remove all the addons" with just one click (e.g. like a reset or similar)

agjohnson: > Though it still seems very odd to me that the hotkeys would keep track of any state.

The only reason we need to keep track of state is because our hotkey implementation is toggling. Hotkeys is unique because it doesn't have to show any state in the UI.

All other UI built over diffing will not use a toggle event. These UI elements will need to show the state of diffing.

We can add a toggle event, but it's only useful for this particular case. We do need the other two events still. I think it's not worth making a new pattern for just this.

ericholscher: How would the hotkeys toggle if we don't have a toggle event, then? The hotkeys can't keep track of state as well as the Addon -- could the hotkeys query the state from the addon, and then trigger the right event?

agjohnson: What I described above would support toggling. It involves keeping track of the last state event, but this is similar to how any UI element will have to store this state.

That is, a diffing dropdown or a slider UI will need to store what the current state is in order to detect when the value changes. These UI would listen for the show/hide events and update the dropdown/slider value to the new selected option.

In the case of hotkeys, storing the selected option is just a variable instead of a dropdown select option.

> could the hotkeys query the state from the addon

That would get into issues with polling -- the UI and hotkeys need to continually poll for diffing addon state changes.

Evented patterns are already doing this, but skipping the polling. We're executing "when the diffing state == showing, execute this callback updating this element", which looks quite like "ask the diffing addon for it's state and then run code updating this element based on the response".

This is where a toggle event doesn't quite fit though. Other addons can't listen for a toggle event to update their UI. It doesn't indicate what the state of diffing is, only that the state changed.

humitos: I understand what Anthony says here and I think he is correct.

_However_, my proposal still works in this scenario _only_ because there isn't currently _another_ addon that needs to know the state of the DocDiff.

Once we add another addon that needs to know the state of the DocDiff (eg. UI element as he mentioned), toggle event will break --or, those addons will need to add a listener to the toggle event as well.

agjohnson: > However, I understand my proposal still works in this scenario only because there isn't currently another addon that needs to know the state of the DocDiff.

Yeah exactly. While it does work for this case, a toggle event likely only fits this one case. But, also, the current pattern does work here already as well.

> -or, those addons will need to add a listener to the toggle event as well.

Well I guess what I meant is that a toggle event doesn't work as a listener event for these UI elements either way because they won't be able to tell if the toggle event means showing toggling to hiding or hiding toggling to showing. Show/hide events already describe this state and state change.

ericholscher: I understand the idea here. I do think keeping track of the state of the Addon in multiple places is likely going to lead to bugs where the states get out of sync. I'd really love a way to just have the state stored in 1 place, and have the other Addons query that Addon for state.

I've pushed up a PR with a basic attempt at this, but it doesn't work because the state is never getting toggled off. I'm sure there's something silly I'm doing, but I couldn't figure it out immediately.

src/hotkeys.js
```diff
@@ -89,6 +88,18 @@ export class HotKeysElement extends LitElement {
   connectedCallback() {
     super.connectedCallback();
     document.addEventListener("keydown", this._handleKeydown);
+
+    document.addEventListener(
+      EVENT_READTHEDOCS_DOCDIFF_ADDED_REMOVED_SHOW,
+      function (event) {
+        console.log("DD Show");
+        this.docDiffEnabled = true;
+      },
```
agjohnson, line 97: I'm guessing this is part of your problem: instead of an arrow function, you've used a `function` expression, which _rebinds `this`_. Instead of being the instance of `HotKeysElemeent`, `this` will refer to the caller of the event, likely the document instance.

You should use an arrow function (`(event) => { ... }`) instead. Using this syntax, the inner `this` is still bound to the instance of the outer `HotKeysElement` -- `this` is _not_ rebound in arrow functions.

ericholscher:
This should be working now 🎉

agjohnson (approved):
Updates look good here :+1: 

Tests might be a good next addition, I'd tackle those in a separate PR though.

src/docdiff.js
```diff
@@ -129,42 +130,56 @@ export class DocDiffElement extends LitElement {
...
+      new_body,
+      VISUAL_DIFF_OPTIONS,
+    );
+    new_body.replaceWith(diffNode.firstElementChild);
+  }
+
   enableDocDiff() {
-    if (this.config === null) {
+    if (this.config === null || this.enabled) {
+      console.debug("Ignoring enableDocDiff: it was already enabled");
```
humitos, line 182: This message is not correct when `this.config === null`. It will be confusing in that case since it won't be enabled because there is no config.

src/hotkeys.js
```diff
@@ -60,12 +58,13 @@ export class HotKeysElement extends LitElement {
...
-      if (this.docDiffShowed) {
+      if (this.docDiffEnabled) {
         event = new CustomEvent(EVENT_READTHEDOCS_DOCDIFF_HIDE);
-        this.docDiffShowed = false;
       } else {
         event = new CustomEvent(EVENT_READTHEDOCS_DOCDIFF_ADDED_REMOVED_SHOW);
-        this.docDiffShowed = true;
+      }
+      if (event !== undefined) {
+        document.dispatchEvent(event);
```
humitos, line 67: We should remove this from here. The event is already dispatched in line 82 and it also prevent its default.

humitos (commented):
Looks good with some small modifications.

src/docdiff.js
ericholscher: Why would it ever get called without a config?

humitos: Why are we checking for `this.config === null` here?

ericholscher: Not sure. It was in the code originally, but there's a lot of the flow that I don't fully understand.

humitos: Well, then you can leave it and move the `console.debug` to be called _only_ on `this.enabled` then.

src/docdiff.js
```diff
@@ -121,13 +121,15 @@ export class DocDiffElement extends LitElement {
...
 
-  handleClick(e) {
-    if (e.target.checked) {
-      this.enableDocDiff();
-    } else {
-      this.disableDocDiff();
-    }
-  }
+  // This code isn't used until we show a UI,
+  // and even then we'll want to trigger events to match state?
```
humitos, line 125: Ah, this code was used in the `render()` method from above originally when I tried using a slider. We never deployed that, so we can probably remove that and this chunk as well.

humitos (approved):
Good work! 🚀 💯

Issue #88, opened 2023-08-10
Notification: show build tip/warning messages on PR builds


humitos:
It would be really good to find a way to show build "tip" and "warning" notifications **on the rendered version of the documentation**. This will make people reading the PR version of that documentation to be aware of potential issues on that particular build.

We could:

- show a small 🔔 icon somewhere in the page that lists all the notification
- show a notification (similar to the current one) below/above the current one with the notification we want to communicate (the background color could 🟡 or 🟦 depending if it's a warning or a tip

### Example

This notification shown in the build's detail page (that most people don't see)

![Screenshot_2023-08-10_17-59-23](https://github.com/readthedocs/addons/assets/244656/29336887-72ec-4955-8642-cffd45e809a9)

could be shown **in the rendered version** as:

![Screenshot_2023-08-10_17-59-01](https://github.com/readthedocs/addons/assets/244656/4d368e16-d7b1-450a-b138-1a4937a5249a)

Related:
* https://github.com/readthedocs/readthedocs.org/issues/10598

humitos:
> exposing a "build status" addon that just shows a little 🟢 🟡 🔴 status in the doc page

I was thinking that if it's _just the build status_, we could include an UI element inside the "This page was created from a pull request build" notification and keep the notification _and_ the status together in the same notification, since they are pretty related each other.

ericholscher:
Yea, I think there's a few ways to show the data. I don't feel strongly about the UX, but a simple initial version would be neat just as a proof of concept, then we can build from there.

humitos:
I started playing with this and found two interesting things:

We had a bug in our code for `builds.currents` when not sending the `url=` parameter. I opened a PR for this at https://github.com/readthedocs/readthedocs.org/pull/11068

> exposing a "build status" addon that just shows a little 🟢 🟡 🔴 status in the doc page.

To implement the use case that @ericholscher is describing here we don't have to cache the addons API responses for any page on PR builds. This is because the response will be _dynamic_ and will depend on the _ongoing_ build. This may not be a problem since we expect to have very little requests from PRs and we expect the cache to be invalidated frequently (on each push), but it's something to keep in mind if we decide to move forward with the implementation of your use case.

ericholscher:
Good point on the caching. It might be a use case for an additional API call for that addon if it's enabled or something -- but I agree, it's probably fine to just not cache PR builds in this case since they are hit _way_ less often than production.

humitos:
[...]

PR #446, opened 2024-11-21
Draft: Flyout: Build status indicator

ericholscher:
Do a very basic build status indictor,
which shows the status of the current/last build.

I couldn't get it to center vertically for some reason,
but just wanted to throw something up as a starting point. 

## Example 

Screenshot 2024-11-21 at 3 25 59 PM

### DIFF
--- a/src/flyout.css
+++ b/src/flyout.css
@@ -84,11 +84,34 @@ header > span svg.icon {
   vertical-align: middle;
 }
 
+header > span.build-status {
+  width: var(
+    --readthedocs-flyout-header-font-size,
+    calc(var(--addons-flyout-font-size) * 1.125)
+  );
+}
+
 header > img.logo {
   text-align: center;
   max-width: 106px;
   max-height: 22px;
   margin-right: 10px;
+}
+
+.build-status-gray {
+  color: gray;
+}
+
+.build-status-green {
+  color: green;
+}
+
+.build-status-red {
+  color: red;
+}
+
+.build-status-yellow {
+  color: yellow;
 }
 
 main {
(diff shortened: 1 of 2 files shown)


agjohnson:
Don't stop exploring addons and JS, but I do think we should discuss this feature first.

This will show failed build state to all reader users. That seems bad for large projects and commercial users, where reader users most likely don't know what Read the Docs or a failed build is. A failed state, red circle, or describing the build as failed all seem like things we should only throw at maintainers.

I'd say the flyout is the wrong place to surface this, and this would make more sense in a dedicated PR build UI. This might be the same UI as diffing too.

humitos:
> Don't stop exploring addons and JS, but I do think we should discuss this feature first.

I agree on this. We touched on this idea in a previous issue but we found we need to do something else regarding caching: https://github.com/readthedocs/addons/issues/88

> I'd say the flyout is the wrong place to surface this, and this would make more sense in a dedicated PR build UI. This might be the same UI as diffing too.

We talked about maybe adding an indicator on the notification.

I think we are entering on the phase were we would get a lot of benefit from discussing "what are the next steps regarding addons UI" as soon as possible. We all want to implement more addons features, but nobody feels comfortable about _where_ to put them. Building these features will be pretty quick once we define that.

ericholscher:
> I agree on this. We touched on this idea in a previous issue but we found we need to do something else regarding caching: https://github.com/readthedocs/addons/issues/88

This is useful context. I see we already noted in that issue that we'll want to hit the API directly for this data to avoid caching, which makes a lot of sense :+1:

humitos:
I'm closing this PR for now. We can come back to it if we still consider it valid. However, I think this won't happen before #574 -- We should probably consider it there directly.

Issue #311, opened 2024-05-21
EthicalAd: implement better positioning


humitos:
We wrote the logic we want to implement in https://github.com/readthedocs/addons/pull/295#discussion_r1600375886

> Placement preference for RTD ad injection:
>
> Using stickybox image (These float if the screen is wide, automatically):
> * Sidebar like content (`role=sidebar`) (Currently we check for sidebar height, and if it's too long, we put it in the footer -- but this is probably not necessary?)
> * Footer like content (`role=footer`)
> * Main body content (`role=main`)
>
> Using text:
> * Fixed footer (Not conflicting with the flyout?)

It would be good to review it again when polishing the ad integration.

humitos:
After the initial deploy from yesterday, we may reconsider this and avoid showing an ad if it doesn't fit in the screen in a well known position.

ericholscher:
Talking about this at the offsite, I'd like to look at redoing something similar to our old footer ad on RTD, and basically just appending a text ad to the bottom of the detected content with our `main content` selector. 

We could add a placement id to these, and track them to see how many views it's getting over time. If we find themes that are getting a lot of traffic, we go in and work to implement a better default placement for that theme/tool.

PR #447, opened 2024-11-22
EthicalAds: Default to fixedfooter for smaller screens

ericholscher:
This is something I've wanted for a little while.
I don't think we're quite ready to ship it,
but I think it makes sense in a lot of cases.

I'm generally becoming more excited about this placement,
especially when the floating one is pretty intrusive.

* Closes https://github.com/readthedocs/addons/issues/311

## Demo

Screenshot 2025-01-14 at 4 12 54 PM
Screenshot 2025-01-14 at 4 13 09 PM
Screenshot 2025-01-14 at 4 26 14 PM
Screenshot 2025-01-14 at 4 26 21 PM

 
Screenshot 2025-01-14 at 5 02 37 PM


ericholscher:
Marking this ready for review after our previous conversation. I'd like to see us doing the fixed footer by default more, since I think it's a much more user friendly ad format. We're hopefully gonna work on a nice fixed footer/header format with a small image, which I think will be even better, but I think this is a nicer UX for default.

In general I really dislike how not integrated the ad feels on non-Sphinx themes, so I think this is a "gentler" approach to ads that isn't quite as impactful for users who want to move to RTD for non-Sphinx themes.

ericholscher:
Looks like this hides the "sidebar" toggle on docify: https://test-builds.readthedocs.io/en/docsify/#/ -- and there wasn't an element I could find to add margin-bottom to, because I think they're fixed positioning....

The struggle of placement on arbitrary content :(

humitos:
> In general I really dislike how not integrated the ad feels on non-Sphinx themes, so I think this is a "gentler" approach to ads that isn't quite as impactful for users who want to move to RTD for non-Sphinx themes.

Yeah, I think we should keep adding more frameworks to our heuristic detection and create better ad placement for specific documentation tools. Sphinx is well integrated because we did the work there. The same happens with all the other we have listed at: https://github.com/readthedocs/addons/blob/8ccaa33fab84037e880f0ae29ff565c40eb597de/src/ethicalads.js#L73-L142

We can do the same for Markdoc, mdbook and similars to have better integrations on those as well. It shouldn't be hard. We need to find the right CSS selector.

I'm not opposed to have a better default for frameworks we don't know; but we for those we are aware they exist and people are using them, we should integrate them better by using custom positioning.

> NOTE: This is somewhere that it would be interesting to have a feature flag, so that we could test it a bit before rolling it out more widely.

As I mentioned in the meeting, I would not use feature flags for this, but use a `AddonsConfig.ethicalads_*` field instead.

humitos:
I added an example of what I'm saying in https://github.com/readthedocs/addons/pull/470. You can see it's pretty easy to create these custom positioning based on the framework. I think they will always work better (less intrusive, nicer look&feel) than a generic default.

humitos:
> I'm generally becoming more excited about this placement,
> especially when the floating one is pretty intrusive.

By they way, when performing visual changes, I would really appreciate having a screenshot of how it looks --that way is a lot easier to do a review; since it doesn't require me to clone the PR and testing it locally.

src/ethicalads.js
```diff
@@ -158,6 +158,20 @@ export class EthicalAdsAddon extends AddonBase {
           placement,
           elementInsertBefore.lastChild,
         );
+      } else if (window.innerWidth <= 1300 && window.innerWidth > 768) {
```
agjohnson, line 161: I just noticed this logic and the same above. These conditionals are only evaluated once, so don't catch the window resizing or rotating. Users have to reload before the placement style would shift in this case. These widths might be better off as CSS media queries, which are reactive.

agjohnson (commented):
@humitos point on per-tool styling is probably better overall, but just noting a couple things with the approach here.

ericholscher:
Next steps:

* Figure out a somewhat reliable way to bump most content "up"
* Ship fixed footer as default instead of floating image ad
* Continue building better integrations into various themes/tools as they come up

ericholscher:
I think this PR should be ready to be merged. I'm adding a class to the main content node to add a bottom margin so hopefully the ad doesn't cover the content. 

The other option I was wondering about is adding a close button to the ad, just in case the content under it needs to be accessed. @davidfischer thoughts on building that into the placement itself, or should we customize it?

humitos:
@ericholscher can you post screenshots of how it looks under different doc tools? That would help with the review.

ericholscher:
Screenshot 2025-01-14 at 4 12 54 PM
Screenshot 2025-01-14 at 4 13 09 PM
Screenshot 2025-01-14 at 4 26 14 PM
Screenshot 2025-01-14 at 4 26 21 PM

src/ethicalads.css
```diff
@@ -128,56 +128,8 @@ div.ethical-footer {
...
-  text-decoration: none;
-}
-.ethical-fixedfooter .ethical-close {
-  position: absolute;
-  top: 0;
-  right: 5px;
-  font-size: 20px;
-  line-height: 20px;
+.ethical-fixedfooter-margin {
+  margin-bottom: 5em;
```
ericholscher, line 132: @agjohnson Is there a way to have this match the height of the ad instead of a random height? 

Seems like you could hack it with something like:

```
// Get the height of the target element
const targetElement = document.querySelector('.target-class');
const matchingElement = document.querySelector('.matching-class');

// Match the height dynamically
matchingElement.style.height = `${targetElement.offsetHeight}px`;
```

but I don't think there's a way to do it non-dynamically, right?

humitos: Even calculating the margin at load time, it will break when resizing the window.

I think the problem here is `position: fixed` (that comes from EthicalAd client CSS). Using a more selector with more specificity (😉 @agjohnson ) we can override it. Example showing how it works:

```css
#readthedocs-ea {
  position: static;
}
```

With this, we don't need to add the extra class `.ethical-fixedfooter-margin` to all the elements as you are doing in https://github.com/readthedocs/addons/pull/447/files#diff-10e47ee23788d0e8d3d4e13cee726fd04c8fe035922a74ac1c7d432db1dfb9faR191

I understand this works because we are adding this element as the last one in the `document.body`.

However, it does not work well in all the documentation tools because the ad may end behind a scroll 😞 

## Demo

### Material for MkDocs

Peek 2025-01-15 10-51

### Sphinx Furo

Peek 2025-01-15 10-58

src/ethicalads.js
```diff
@@ -158,13 +158,41 @@ export class EthicalAdsAddon extends AddonBase {
...
+        const root_node = document.querySelector(docTool.getRootSelector());
+        const footers = document.querySelectorAll("footer, .footer");
+
+        if (root_node) {
+          root_node.classList.add("ethical-fixedfooter-margin");
+        }
+
+        footers.forEach((footer) => {
+          footer.classList.add("ethical-fixedfooter-margin");
+        });
```
humitos, line 193: This could end up adding a lot of extra space at the bottom depending on the number of elements the page has.

### Using stickybox

Screenshot_2025-01-15_11-03-26

### Using fixedfooter

Screenshot_2025-01-15_11-03-39

src/ethicalads.js
```diff
@@ -158,13 +158,41 @@ export class EthicalAdsAddon extends AddonBase {
...
+
+        // Inject the actual ad
```
humitos, line 195: It seems than depending on the theme/doctool the may be injected in a weird place and not covering the full width:

### Solidity

Screenshot_2025-01-15_11-10-41

### Our docs

Screenshot_2025-01-15_11-10-20

humitos (commented):
I think the idea is good, but it's hard to do it correctly in a reliable way. I don't think it has to be _perfect_ [^1], but at least I would like to fix those cases where it seems to be broken (not using the full width of the page and adding a lot of blank space at the bottom).

With those changes, I think I will feel more comfortable deploying this and keeping an eye on those new cases where it breaks to fix them as we go.

This change will affect _all the documentation pages_ which looks like a big change to deploy something we don't feel comfortable about. We talked about adding `AddonsConfig.ethicalads_type = "text"` and add it a few projects to start. That may be still an option as well.

[^1]: making it _reactive_ could be a v2 of this implementation, for example

src/ethicalads.css
agjohnson: There isn't pure CSS solution here given the fixed footer is a separate element from the elements you're adding classes to for the margin -- there isn't a way to reference between elements like this unless one element is a direct descendant. In this case you have more options and `calc()` might be able to get there.

I'd avoid a JS solution here especially using `Element.style` -- my note above applies here.

It feels like we should have multiple explicit styles by viewport size, not necessarily try to do this dynamically -- that is we'd have explicit heights for mobile, table, desktop widths.

src/ethicalads.js
```diff
@@ -158,13 +158,41 @@ export class EthicalAdsAddon extends AddonBase {
...
+        //   placement.setAttribute("data-ea-style", "stickybox");
+        //   this.addEaPlacementToElement(placement);
+        //   // `document.body` here is not too much relevant, since we are going to
+        //   // use this selector only for a floating stickybox ad
+        //   const elementInsertBefore = document.body;
+        //   elementInsertBefore.insertBefore(
+        //     placement,
+        //     elementInsertBefore.lastChild,
+        //   );
+      } else if (window.innerWidth > 768) {
```
agjohnson, line 176: This is a case for `matchMedia`: https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia

Noted earlier that this check isn't dynamic currently, it's only evaluated once. Using `matchMedia` as an event gives a way to do this after page load.

agjohnson (commented):
Noted some options here, but I'd overall agree that the fixed footer is going to be the hardest placement to get correct across all tools/themes. Text ad would indeed be an easier place to start for a site wide default placement.

This seems like it might be close but it does need some more precision. Altering users styles without precision is bound to cause some friction or confusion with users.

src/ethicalads.js
humitos: > Do we want it full width of the frame or not?

I think it should be full width, yes.

src/ethicalads.js
ericholscher: We can't easily reconfigure the ad without doing another API call, so we don't want it to be dynamic. We want to have a single ad load on page load, and stick with that, otherwise it will get weird.

src/ethicalads.js
ericholscher: It shouldn't be showing on the RTD theme, since that has a custom placement. I'm not really sure what the change being suggested here is -- isn't the ad getting injected into the document.body?

src/ethicalads.js
ericholscher: Should we just not try and just the bottom margins at all? The issue is that it covers up fixed position content on other themes (eg. mkdocs-material), so I wasn't sure what the right approach was.

ericholscher (commented):
I'm not really sure what is being suggested here, but feel free to update this PR with an approach that y'all think will work.

src/ethicalads.js
humitos: The ad is being injected on RTD theme here. I think that happens because of the current logic: "if the specific placement it not above the fold, we show the default placement". The "default placement" now is fixedfooter, and that's why it's being shown.

src/ethicalads.js
humitos: If I had to choose, I would prefer to cover footer content than breaking layout (adding weird margins). In either case, we can work on "better placement for that particular doctool" when we see it or when users complain 😄

humitos: We already have a specific positioning for Material for MkDocs that's working already: https://docs.pypi.org/

ericholscher: Right, but not for smaller screen sizes, like we are discussing?

PR #448, opened 2024-11-22
Add `article` to the FALLBACK_DOCTOOL list

ericholscher:
It seems this is common enough in other doc tools,
I think we should probably include it in the default?

I hit this with markdoc, where `main` includes the sidebars and other non-primary content.


agjohnson (approved):
Yeah agreed. This seems to be an oversight from splitting up the selectors by tool, I think we already wanted this.

PR #449, opened 2024-11-23
Add initial Command Pallette addon

ericholscher:
This is one I'm pretty excited about.
It takes the Command Pallette pattern from Slack, GitHub, etc.,
and adds it to any documentation page.

I think this could be an interesting UX for our future UX,
and enable a lot of really interesting power user features.

The really nice thing is that it's totally based on a keyboard shortcut,
so doesn't require any UX, and generally folks won't want to disable it.

### DIFF
--- a/src/commands.js
+++ b/src/commands.js
@@ -1,0 +1,506 @@
+import { library, icon } from "@fortawesome/fontawesome-svg-core";
+import {
+  faCircleXmark,
+  faMagnifyingGlass,
+  faCircleNotch,
+  faBinoculars,
+  faBarsStaggered,
+} from "@fortawesome/free-solid-svg-icons";
+import READTHEDOCS_LOGO from "./images/logo-wordmark-dark.svg";
+
+import styleSheet from "./search.css";
+import { AddonBase, addUtmParameters } from "./utils";
+import {
+  EVENT_READTHEDOCS_COMMANDS_SHOW,
+  EVENT_READTHEDOCS_COMMANDS_HIDE,
+  EVENT_READTHEDOCS_DOCDIFF_ADDED_REMOVED_SHOW,
+  EVENT_READTHEDOCS_DOCDIFF_HIDE,
+  EVENT_READTHEDOCS_SEARCH_SHOW,
+  EVENT_READTHEDOCS_SEARCH_HIDE,
+  EVENT_READTHEDOCS_FLYOUT_SHOW,
+  EVENT_READTHEDOCS_FLYOUT_HIDE,
+} from "./events";
+import { html, nothing, LitElement } from "lit";
+import { classMap } from "lit/directives/class-map.js";
+
+const MIN_CHARACTERS_QUERY = 3;
+
+export class CommandPaletteElement extends LitElement {
+  static elementName = "readthedocs-commands";
+
+  static properties = {
+    config: {
+      state: true,
+    },
+    show: { state: true },
+    inputIcon: { state: true },
+    commands: { state: true },
+    commandResults: { state: true },
+    cssFormFocusClasses: { state: true },
+    triggerKeycode: { type: Number, attribute: "trigger-keycode" },
+    triggerSelector: { type: String, attribute: "trigger-selector" },
+    triggerEvent: { type: String, attribute: "trigger-event" },
+    htmlResults: { state: true },
+  };
+
+  static styles = styleSheet;
+
+  constructor() {
+    super();
+
+    library.add(faMagnifyingGlass);
+    library.add(faCircleNotch);
+    library.add(faBarsStaggered);
+    library.add(faCircleXmark);
+
+    this.config = null;
+    this.show = false;
+    this.cssFormFocusClasses = {};
+    this.commands = [
+      {
+        path: "Show DocDiff",
+        sections: [],
+        action: () => {
+          const event = new CustomEvent(
+            EVENT_READTHEDOCS_DOCDIFF_ADDED_REMOVED_SHOW,
+          );
+          document.dispatchEvent(event);
+        },
+      },
+      {
+        path: "Hide DocDiff",
+        sections: [],
+        action: () => {
+          const event = new CustomEvent(EVENT_READTHEDOCS_DOCDIFF_HIDE);
+          document.dispatchEvent(event);
+        },
+      },
+      {
+        path: "Show Search",
+        sections: [],
+        action: () => {
+          const event = new CustomEvent(EVENT_READTHEDOCS_SEARCH_SHOW);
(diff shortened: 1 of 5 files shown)


humitos:
I think this is an interesting idea to explore more 👍🏼 . It seems we don't need to create a different addon and UI for this and it can probably be integrated with the search addon -- like a feature of the search addon: you type anything and you see the results for the "commands" and for the "search" in the same UI.

ericholscher:
@humitos any additional thoughts here ^^

I still think the Command Pallette is a powerful UX for exposing a bunch more functionality to users, and would love to get at least a internal-only test going at some point soonish.

src/commands.js
```diff
@@ -0,0 +1,566 @@
...
+  firstUpdated() {
+    this.className = this.className || "raised floating";
+  }
+
+  render() {
+    if (this.config === null) {
+      console.log("Config is null, rendering nothing");
+      return nothing;
+    }
+    console.log("Rendering command palette");
```
humitos, line 208: I'd remove all these `console.log`s before merging this PR.

src/commands.js
```diff
@@ -0,0 +1,566 @@
...
+    return this.renderCommandPalette();
+  }
+
+  renderCommandPalette() {
+    console.log("Rendering command palette modal");
+    return html`
+      <div ?hidden=${!this.show} role="search">
```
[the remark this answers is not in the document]
humitos: Note there are more references to search in other places as well.

src/commands.js
```diff
@@ -0,0 +1,566 @@
...
+    // Convert query to lowercase for case insensitive comparison
+    const lowerCaseQuery = query.toLowerCase();
+
+    // Simulate fetching results from in-memory data with fuzzy searching
+    const filteredResults = results.filter((result) => {
+      const pathMatch = result.title.toLowerCase().includes(lowerCaseQuery);
+      const sectionMatch = result.sections.some((section) =>
+        section.title.toLowerCase().includes(lowerCaseQuery),
+      );
+      return pathMatch || sectionMatch;
```
humitos, line 367: Does this `||` combines both lists and returns the resulting of that combination?

src/commands.js
```diff
@@ -0,0 +1,566 @@
...
+    );
+
+    super.disconnectedCallback();
+  }
+}
+
+export class CommandsAddon extends AddonBase {
+  static jsonValidationURI =
+    "http://v1.schemas.readthedocs.org/addons.commands.json";
+  static addonEnabledPath = "addons.filesections.enabled";
```
humitos, line 545: The API should return `addons.commands`. File sections seems to be one of the features of the commands palette, not the addon itself.

```suggestion
  static addonEnabledPath = "addons.commands.enabled";
```

src/data-validation.js
```diff
@@ -274,6 +274,48 @@ const addons_filetreediff = {
...
 };
+// Validator for FileSections Addon
+const addons_filesections = {
+  $id: "http://v1.schemas.readthedocs.org/addons.commands.json",
+  type: "object",
+  required: ["addons"],
+  properties: {
+    addons: {
+      type: "object",
+      required: ["filesections"],
```
humitos, line 285: This shouldn't be required since projects without `filesections` won't have access to the commands palette.

src/data-validation.js
```diff
@@ -274,6 +274,48 @@ const addons_filetreediff = {
     },
   },
 };
+// Validator for FileSections Addon
+const addons_filesections = {
+  $id: "http://v1.schemas.readthedocs.org/addons.commands.json",
+  type: "object",
+  required: ["addons"],
+  properties: {
```
humitos, line 282: Here you want to have `projects`, `versions` and `builds` required as well.

src/data-validation.js
```diff
@@ -274,6 +274,48 @@ const addons_filetreediff = {
     },
   },
 };
+// Validator for FileSections Addon
+const addons_filesections = {
```
humitos, line 278: This should make reference to commands, instead of filesections.

src/hotkeys.js
```diff
@@ -43,6 +44,7 @@ export class HotKeysElement extends LitElement {
     this.docDiffShowed = false;
 
     this.searchHotKeyEnabled = this.config.addons.hotkeys.search.enabled;
+    this.commandPaletteHotKeyEnabled = true;
```
humitos, line 47: We should decide for one name and stick with it everywhere in the code; variables, functions and event names: `Command` or `CommandPalette`.

src/hotkeys.js
```diff
@@ -75,11 +78,25 @@ export class HotKeysElement extends LitElement {
...
+      document.activeElement.tagName !== "READTHEDOCS-COMMANDS"
     ) {
       event = new CustomEvent(EVENT_READTHEDOCS_SEARCH_SHOW);
     }
 
+    // Command Palette
+    if (
+      this.commandPaletteHotKeyEnabled &&
+      (e.metaKey || e.ctrlKey) &&
+      e.key === "k" &&
```
humitos, line 91: This hotkey combination should come from the API.

humitos (commented):
I did a code review and I left a few comments and suggestions.

I think this is an interesting concept. However, I'd say we should work on other pieces before moving forward with this (eg. backend API for `.json` file, add required AddonsConfig fields, etc).

This implementation copies a lot of the search code/UI which, unfortunately, is the least polished one. I mentioned in another comment that I'd like to see the Command Palette and Search UI combined, making these command items inside the same UI.

I don't think this is ready to be exposed to users yet, but we can merge it for internal-only testing once we have done the backend work if you want.

ericholscher:
> This implementation copies a lot of the search code/UI which, unfortunately, is the least polished one. I mentioned in another comment that I'd like to see the Command Palette and Search UI combined, making these command items inside the same UI.

This is what I was asking for feedback on, was my discussion about this above: https://github.com/readthedocs/addons/pull/449#issuecomment-2498538315

In particular, how does it make sense to combine things that are realtime lookups with search lookups that are delayed. I think that's going to be a much harder UX challenge. 

I do think the Command Palette concept could be a pretty powerful thing to expose, and give us a bunch more functionality that we can add outside of the Flyout, but mostly for advanced users. I think if we shipped it, I would use it on our docs everyday tho, which I think is a pretty powerful vote for it being something powerful.

humitos:
> In particular, how does it make sense to combine things that are realtime lookups with search lookups that are delayed. I think that's going to be a much harder UX challenge.

We can show the results from commands immediately at the top, and below them in another section, the results that comes from the search API. Long search queries won't have any result from commands so that section won't be shown in those cases. We can distinct these sections/results visually somehow.

PR #459, opened 2024-12-02
Test: add a few test for `DocumentationTool` helper

### LINKED ISSUE #445: Tests: add tests for `DocumentationTools` class
We need to add tests for all the heuristic we are using in this class to detect the documentation tool and return the CSS selectors once the pattern is more settled down.

humitos:
Closes #445


agjohnson (approved):
:100:

PR #462, opened 2024-12-05
Visual diff: enable it via URL parameter by triggering an event

### LINKED ISSUE #461: When `?readthedocs-diff=true` we can't disable visual diff
Steps to reproduce it:

1. https://docs--11751.org.readthedocs.build/en/11751/user-defined-redirects.html?readthedocs-diff=true
2. Press `d`

Visual diff should be disabled, but it doesn't --another layer of diff is added somehow.

humitos:
We trigger the event here so any other addon that requires to keep the state of it can perform their corresponding action as well.

Hotkeys addons needs to make change `this.docDiffShowed = true` when the URL argument is present, otherwise it's not possible to disable it.

Closes #461


ericholscher:
Isnt this already fixed in https://github.com/readthedocs/addons/pull/441/ ?

humitos:
Ah, yes, maybe 😁. I'll double check once that PR gets merged.

humitos:
Closing as Eric's PR already solves this issue.

Issue #157, opened 2023-09-26
Support Single Page Applications (SPA) for addons that depend on the URL


humitos:
Material for MkDocs and other doctools (I think that Docusaurus too) implement a feature that load the content of the page _without_ making a regular request to the server [^1]:

> When instant loading is enabled, clicks on all internal links will be intercepted and dispatched via [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) without fully reloading the page.

This makes addons that depend on the actual URL to not work as expected (e.g. docdiff). When loading the page, the addons API is hit with `url=` but then the user clicks on an internal link, which uses the instant reload feature to load the page content but our addons doesn't do another request with the `url=` field updated, making docdiff to use the old URL to compare against.

I think we could listen to an event to check if the URL has changed and if so, make a request to the addons API again with the updated `url=` attribute. It seems there is a navigator API to do this, but it's experimental for now, tho: https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API

[^1]: https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#instant-loading

humitos:
Besides DocDiff, now that we support "keeping on the same page when changing version/language" in the Flyout, this addon also suffers this problem. This happens on Docusaurus and Markdoc too.

zanderle:
@humitos so if I understand correctly, you want to use Navigation API? The support for this isn't looking to good yet :/ https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate_event

An alternative (which I'm not a fan of) would be to monkey-patch `history.pushState`, so that it would fire up `PAGE_CHANGED` every time it's called.

Another possibility (not sure how feasible this would be) could be to add a `script` to every page that fires some event that addons listen to and act accordingly?

OR could we make this a responsibility of the theme instead? If your theme is using SPA, you need to fire `ADDONS_PAGE_CHANGED` event, to make addons work correctly?

humitos:
> @humitos so if I understand correctly, you want to use Navigation API? The support for this isn't looking to good yet :/ [developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate_event](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate_event)

Hrm, OK. That sounds like a no go, then. Support is pretty limited 😞 

[...]

PR #465, opened 2024-12-05
Visual diff: trigger `DOM_CHANGED` event after enable/disable

### LINKED ISSUE #460: Passing `?readthedocs-diff=true` breaks Links Preview
Steps to reproduce:
1. Go to https://docs--11751.org.readthedocs.build/en/11751/user-defined-redirects.html?readthedocs-diff=true
2. Hover on "custom domains" under "Root URL redirect at /" section
3. No preview is shown, and no error is reported

If you go to https://docs--11751.org.readthedocs.build/en/11751/user-defined-redirects.html (without `readthedocs-diff`), links preview works properly.

humitos:
A new event `READTHEDOCS_ROOT_DOM_CHANGED` is triggered when the "Visual diff" is enabled/disabled after the DOM was modified.

This allows other addons to subscribe to this event to re-initialize if required. I used this event from links preview to call `setupTooltips` again to re-install tooltips on these links.

I migrated links preview to a `LitElement` to be able to make usage of this pattern and follow what we already had.

We will be able to do something similar on #157

Closes #460


src/events.js
```diff
@@ -9,6 +9,8 @@ export const EVENT_READTHEDOCS_FLYOUT_SHOW = "readthedocs-flyout-show";
 export const EVENT_READTHEDOCS_FLYOUT_HIDE = "readthedocs-flyout-hide";
 export const EVENT_READTHEDOCS_ADDONS_DATA_READY =
   "readthedocs-addons-data-ready";
+export const EVENT_READTHEDOCS_ROOT_DOM_CHANGED =
```
ericholscher, line 12: Should we have a docstring or similar to describe the events here?

ericholscher (approved):
Looks like a nice refactor to me.

src/events.js
humitos: Added 👍🏼

PR #468, opened 2024-12-09
API: use APIv3 endpoint for resources

### LINKED ISSUE #356: API: consider using APIv3 standard endpoints
Now that we are allowing public access to APIv3 endpoints, we can think about making different API queries using the common APIv3 endpoints to fetch small chunks of data instead doing one big call to the specific `/_/addons/` one.

This would be transparent for our own JavaScript library and for users of the `CustomEvent` because it will be managed internally and the _exact_ same JSON-like object will be returned.

As a example,
- `projects.current` field can be populated with https://docs.readthedocs.io/en/stable/api/v3.html#project-details
- `versions.current` field with https://docs.readthedocs.io/en/stable/api/v3.html#version-detail 
- `versions.active` field with https://docs.readthedocs.io/en/stable/api/v3.html#versions-listing

This may be something good to explore since I'm seeing some benefits on caching (not all of the responses will be purged on each new build) and also in maintenance as well, since it's going to be one less big endpoint to maintain over time. Besides, we will become users of our own APIv3 which will make us to keep improving it over time.

* https://github.com/readthedocs/readthedocs.org/pull/11485

humitos:
Initial implementation as a POC to give it a try. It works fine at first sight and I think it could be a good idea to move forward with it. I didn't find any blocker yet.

Closes #356
Requires https://github.com/readthedocs/readthedocs.org/pull/11831

### DIFF
--- a/src/readthedocs-config.js
+++ b/src/readthedocs-config.js
@@ -143,11 +143,6 @@ export async function getReadTheDocsConfigUsingAPIv3(sendUrlParam) {
   const versionResponse = fetch(
     addons.readthedocs.urls.api.v3.versions.current,
   );
-  // TODO: the results come paginated, so we are only seeing the first 10 results.
-  // We need to perform more requests to get all the resulsts.
-  // What's the correct way to do that?
-  //
-  // The response comes with a `next` attribute that has the URL for the next 10 results.
   const activeVersionsResponse = fetch(
     addons.readthedocs.urls.api.v3.versions.active,
   );


src/readthedocs-config.js
```diff
@@ -129,6 +129,66 @@ export function getReadTheDocsConfig(sendUrlParam) {
...
+  const versionResponse = fetch(
+    addons.readthedocs.urls.api.v3.versions.current,
+  );
+  const activeVersionsResponse = fetch(
+    addons.readthedocs.urls.api.v3.versions.active,
+  );
+  const buildResponse = fetch(addons.readthedocs.urls.api.v3.builds.current);
+  const filetreediffResponse = fetch(
+    addons.readthedocs.urls.api.v3.filetreediff,
+  );
```
ericholscher, line 152: I worry a little bit about one of the API calls failing and leading to weird state. Would be good to put a TODO in here probably to do a retry?

humitos: Yes. I need to re-write this code to handle errors as well. Currently, it breaks completely 😅

src/readthedocs-config.js
```diff
@@ -129,6 +129,71 @@ export function getReadTheDocsConfig(sendUrlParam) {
...
+    addons.readthedocs.urls.api.v3.versions.current,
+  );
+  // TODO: the results come paginated, so we are only seeing the first 10 results.
+  // We need to perform more requests to get all the resulsts.
+  // What's the correct way to do that?
+  //
+  // The response comes with a `next` attribute that has the URL for the next 10 results.
+  const activeVersionsResponse = fetch(
+    addons.readthedocs.urls.api.v3.versions.active,
+  );
```
humitos, line 153: @agjohnson any suggestion for this?

humitos: ah, we already have the `?limit=` parameter implemented in the endpoint 💯

src/readthedocs-config.js
agjohnson: What you have is close, though yeah it does need more error handling -- especially using `fetch()` you should check for `resp.okay`.

One thing though is that this is sort of written more like synchronous code. Normally, you'd want to take advantage of the evented code flow more, and instead of doing operations in bulk and `await` on each of these blocks, you'd let the request and encoding and any other operation run in a single promise.

That is, this is executing:

```
fetch(), fetch(), fetch(), fetch()
resp.json(), resp.json(), resp.json(), resp.json()
```

Where a fully evented flow is more like:

```
fetch(), resp.json()
fetch(), resp.json()
fetch(), resp.json()
fetch(), resp.json()
```

The effect can be subtle, but it does allow for early failure if one request or request response is bad.

What you have here overall is good so far, though I think through restructuring to add some error handling and simplifying things, you'll probably end up with a separate async function for a single request and request encoding.

Issue #51, opened 2023-04-13
Customization patterns


agjohnson:
This is the beginning of a discussion of some customization patterns that we might provide theme maintainers and authors customizing the look and feel of their own documentation.

Some top level points:

- Theme customization likely includes disabling some of our elements that we're injecting
- Visual customization needs could be quite minimal and our CSS variables are enough

Here are a few patterns:

##### Element template customization by JavaScript

Most obvious would be allowing customization of the class used for our element:

```js
class MyCustomNotificationElement extends NotificationElement {
  render() {
    return html`
      <div class="some-bootstrap-classes">
        <h1>${title}</h1>
        <p>${body}</p>
      </div>
    `
  }
}

// Pseudo library usage, this pattern doesn't exist yet
readthedocs.NotificationAddon.elementClass = MyCustomNotificationElement
```

But this is heavy on the JS side too, of course.

##### Element template customization by HTML script

Something like `<script type="text/html">...</script>` is really approachable.

There are some injection attack worries here to work through (see the Lit `unsafeHtml` tag/function for some background). This might influence this pattern.

##### Element template customization by inner HTML

Not sure this is an option even, and it likely has the same trouble with unsafe HTML, but:

```html
<readthedocs-notification>
  <div class="bootstrap-classes">
    ${title}
    <p>${body}</p>
  </div>
</readthedocs-notification>
```

##### More granular Lit elements

I'm not even entirely sure on this one. It feels like JS API consumption would be easier.

But, for example, a Sphinx theme want something like this instead of a notification:

```html
<div>
  Version: {{ version }}
  <readthedocs-version if="external">
    <i class="fas fa-sparkles"></i>
    From a pull request!
  </readthedocs-version>
</div>
```

##### No element customization, push to JS API

The above example feels easier as the following pseudo library usage:

```js
readthedocs.on("versionLoaded", (version) => {
  if (version.isOutdated) {
    const elem = document.querySelector('div.version_selector > p');
    // Add icon to elem
    ...
  } else if (version.isExternal) {
    document.querySelector('header').appendChild(...);
  }
});
```

humitos:
[...]

PR #473, opened 2024-12-16
Add styles based on the documentation tool

humitos:
Use our heuristic to detect the documentation tool/theme and add specific `--readthedocs-*` CSS variables based on that for known tools/themes.

Reference: https://github.com/readthedocs/readthedocs.org/pull/11849#issuecomment-2540234995

### DIFF
--- a/src/flyout.js
+++ b/src/flyout.js
@@ -8,7 +8,15 @@ import { classMap } from "lit/directives/class-map.js";
 import { default as objectPath } from "object-path";
 
 import styleSheet from "./flyout.css";
-import { AddonBase, addUtmParameters, getLinkWithFilename } from "./utils";
+import styleMkDocsMaterial from "./flyout.mkdocs.material.css";
+import styleSphinxFuro from "./flyout.sphinx.furo.css";
+import {
+  AddonBase,
+  addUtmParameters,
+  getLinkWithFilename,
+  docTool,
+} from "./utils";
+import { SPHINX, MKDOCS_MATERIAL } from "./constants";
 import {
   EVENT_READTHEDOCS_SEARCH_SHOW,
   EVENT_READTHEDOCS_FLYOUT_HIDE,
@@ -386,6 +394,17 @@ export class FlyoutAddon extends AddonBase {
 
     for (const elem of elems) {
       elem.loadConfig(config);
+    }
+
+    this.addCustomStyle();
+  }
+
+  addCustomStyle() {
+    const doctool = docTool.documentationTool;
+    if (doctool === MKDOCS_MATERIAL) {
+      document.adoptedStyleSheets.push(styleMkDocsMaterial);
+    } else if (doctool == SPHINX && docTool.isSphinxFuroLikeTheme()) {
+      document.adoptedStyleSheets.push(styleSphinxFuro);
     }
   }
 }
--- a/src/flyout.mkdocs.material.css
+++ b/src/flyout.mkdocs.material.css
@@ -1,0 +1,4 @@
+:root {
+  /* Reduce Read the Docs' flyout font a little bit */
+  --readthedocs-flyout-font-size: 0.7rem;
+}
--- a/src/flyout.sphinx.furo.css
+++ b/src/flyout.sphinx.furo.css
@@ -1,0 +1,4 @@
+:root {
+  /* Reduce Read the Docs' flyout font a little bit */
+  --readthedocs-flyout-font-size: 0.75rem;
+}


src/flyout.js
```diff
@@ -387,6 +395,17 @@ export class FlyoutAddon extends AddonBase {
...
+    this.addCustomStyle();
+  }
+
+  addCustomStyle() {
+    const doctool = docTool.documentationTool;
+    if (doctool === MKDOCS_MATERIAL) {
+      document.adoptedStyleSheets.push(styleMkDocsMaterial);
+    } else if (doctool == SPHINX && docTool.isSphinxFuroLikeTheme()) {
+      document.adoptedStyleSheets.push(styleSphinxFuro);
+    }
```
agjohnson, line 408: This feels like an odd pattern, especially applying rules to `:root` from a shadow dom element. Normally, these rules would be applied to only this single element, not as a global style rule.

This is maybe a place where a `styles` getter can be used instead, especially if none of the `docTool.*` functions require a bound `this` referencing the current element.

agjohnson (commented):
I believe there is a better pattern here for dynamic styles. The Lit documentation probably suggests using `classMap` on the element, which could be another option. We've talked about avoiding classes for styling though, so I mostly lean towards a dynamic `styles` getter if it works easily.

humitos:
I pushed a commit to what I understood it could be a better pattern for this. Let me know what you think.

src/flyout.js
```diff
@@ -20,6 +26,7 @@ export class FlyoutElement extends LitElement {
 
   static properties = {
     config: { state: true },
+    classes: { state: true, type: Object },
```
agjohnson, line 29: Element already has a `classList`, however my note on using properties over class names is most applicable here.

humitos:
I pushed some changes that use `tool=sphinx` and `tool-theme=furo` attributes on `readthedocs-flyout` to define specific CSS rules based on the documentation tool and theme. I think I like this approach. Feedback welcome.

We will need to do the same for the other addons, where we we want to increase/decrease the font sizes. However, I'm happy to start with flyout for now and grow from there if we find this is the pattern we will stick to.

src/flyout.js
```diff
@@ -31,9 +37,12 @@ export class FlyoutElement extends LitElement {
     super();
 
     this.config = null;
+    this.classes = {};
```
ericholscher, line 40: Seems we're setting this, and then overriding it on line 44?

src/flyout.js
```diff
@@ -310,11 +319,12 @@ export class FlyoutElement extends LitElement {
...
     }
 
-    const classes = { floating: this.floating, container: true };
-    classes[this.position] = true;
-
     return html`
-      <div class=${classMap(classes)}>
+      <div
+        tool="${docTool.documentationTool}"
+        tool-theme="${docTool.documentationTheme}"
```
ericholscher, line 325: Do we not want to prefix these with something? I guess we're controlling it, but I imagine we also want the CSS selectors prefixed, unless it's OK because it's in the shadow DOM?

ericholscher (commented):
I like this approach with the CSS -- it feels really explicit which feels good.

agjohnson:
> Also, we can't do :root[data-readthedocs-tool="docusaurus] because that selector doesn't return anything since we are using shadow DOMs.

If you need to manage these variables _inside_ each addon then this is true, but what I was describing was applying these variables from the _host DOM_, at a global level. This is where users should/will be defining our CSS variables either way.

> I was trying to keep things scoped by addon. That is, from flyout we should be defining only settings that affect the flyout (e.g.--readthedocs-flyout-* variables); without affecting other addons/elements.

I do think isolation makes sense, but we can still do addon isolation of styles/variables at a global level. We don't need to apply `--readthedocs-flyout-font-size` to only `readthedocs-flyout` or just one instance of `<readthedocs-flyout>`, that variable can be applied to `html` and still affect `<readthedocs-flyout>` the same.

Thinking more, we don't need an attribute or class on `html`. It feels like our doc detection rules should happen at a addon configuration/initialization and add the following variables at `html`:

```css
// Styles applied for docusaurus
html {
  --readthedocs-flyout-font-size: 0.9rem;
  --readthedocs-notification-font-size: 0.9rem;
}
```

The easiest change for theme maintainers and users will be to add some CSS like:

```css
// `html body` or `html html` is higher specificity than our rules at `html`.
html body, html html {
  --readthedocs-flyout-font-size: 2rem;
  --readthedocs-notification-font-size: 2rem;
}
```

This doesn't involve editing the doc tool theme HTML, so is the easiest control point.

Defining these variables _inside the shadow DOM like you're describing_ might take precedence over host applied variables though, we should test this. This would make it hard/impossible for this easy control of these variables.

This all gives users/theme maintainers some way to control at least the styles applied by our magic, but it doesn't control or allow opt out from our magic detection. Attributes on each addon does accomplish this but it's rather cumbersome -- maintainers would have to define all our addons in HTML theme `<readthedocs-flyout tool="custom" /><readthedocs-notification tool="custom" />...`.

humitos:
By using selectors like `html { --readthedocs-flyout-font-size: 0.65rem }`, I'm only able to modify things exposed by `--readthedocs-*` variables; which is not enough in all cases. For example, I need to modify `line-height` for some of the themes/doctools and I'm not being able to.

I think that using `--readthedocs-*` is getting pretty complicated to allow _us_ improving the look and feel of the addons while also letting users to modify them. We added them as an initial test to allow _ourselves_ to document how to adjust the sizes of the addons because we noticed they didn't render properly everywhere (mainly on Material for MkDocs). However, this PR implements _the better look and feel/styling_ automatically without user intervention. I'd propose to revert the idea of these `--readthedocs-*` variables and give us full control back over the addons until we figure it out a more generic pattern for customization in #51.

Until then, this PR will give users a better UI/UX anyways since they won't have to do anything on their side. In other words, I'd like to give users the best default experience on their doctools/themes (at least the ones we known) for now (the proposal of this PR) and leave a more polished customization for the future.

ericholscher:
I don't fully understand the implementation details here, but the goal I assume is having good defaults, but letting users be able to override them. If we have to choose only 1, I think good defaults is more important, since the initial experience for the user is most important, and getting that right is going to make adoption much more likely. However, it really seems like we should be able to support both, where we can set a default, and the user can override it?

agjohnson:
> or example, I need to modify line-height for some of the themes/doctools and I'm not being able to.

Line height should almost always be derived from `font-size` though, as a proportional value to the height of the font (`1.2em`) or as a calculated value (`calc()`)

> I think we will end up having one --readthedocs-* variable per addon per property if we follow this pattern

What styles do we need that shouldn't be derived from font-size? This would be helpful to enumerate. The majority of styles seem like be positioning and spacing, both mostly proportional to `font-size`.

Even if there are many other variables required, I still think adding more variables is a good pattern. These variables don't need to map directly to CSS styles directly even.

CSS variables are a benefit because they avoid many of the downsides to JS and HTML customization. More control of CSS styles could render further customization needless even. But also important: a worst case failure of CSS customization is just improper display. JS and HTML customization can throw exceptions if we introduce changes that break user customization.

> I'd propose to revert the idea of these --readthedocs-* variables and give us full control

How do we replace user customization for this PR then, what are users expected to do to override our magic?

So far I don't agree with removal, for the points above.

>  think that using --readthedocs-* is getting pretty complicated to allow us improving the look and feel of the addons while also letting users to modify them.

Agreed, I still think this is completely possible here too. Good defaults are helpful, but we really shouldn't be solely responsible for customization at every tool either. Retaining easy customization points shares this load and doesn't cut out theme maintainers.

humitos:
> However, it really seems like we should be able to support both, where we can set a default, and the user can override it?

Yes, we should be able to support both. However, my proposal is to start with good defaults (this PR) and add support for `--readthedocs-*` variables later due to the complexity I found to support both at this point [^1]. The current support of `--readthedocs-*` is half baked and it's not working in a reliable way, so I'm not worried about removing it for now to allow having better defaults.

With that, we can continue discussing how to support these CSS variables in a better way that works for all the cases, while we give a better UX to our users in the meanwhile.

That said, I agree with the points that Anthony raised in the previous comment, but I'm trying to be more practical now while we debate and learn how to implement it in the near future.

[^1]: maybe is not _super hard_ but I don't have the knowledge to make it work yet. It will require researching and learning from my side at least.

src/flyout.js
humitos: I added the `data-` prefix for now.

humitos:
With the latest changes on this PR, this is how the flyout looks by default on different doctools:

Peek 2025-01-08 10-06


My proposal here is to merge it and open a different issue to explore/research/implement the support for CSS variables together with this approach.

ericholscher:
Next steps:

* @agjohnson to implement his suggestion of Doctool setting top-level CSS variables
* We're :100: to ship this PR after that.
* Document and make official support for CSS variables
* Work on internal docs around when and how to use CSS variables when building implementation

agjohnson:
This PR is what I was describing above, for the most part:

- #484

It reuses the documentation tool/theme attribute approach instead. But these rules are applied at the parent DOM and allow for users to control these values still.

I would not set `line-height` specifically. It should be a derived, relative value as should all of the spacing and position values.

I didn't test this further with all of the different tools/themes, I'm not sure how you're testing these. But the underlying pattern for variable use is the most important part here.

humitos:
> I'm not sure how you're testing these.

1. Build all the versions I want to test locally using `test-builds`
2. Go to https://readthedocs-addons.readthedocs.io/
3. Click the slider to make all the links point to my local instance
4. Click all the links and take a look at the addon

humitos:
This looks great to me! It's _almost_ perfect!

Peek 2025-01-09 11-25

humitos:
With this implementation, we can now use a CSS more specific selector to re-define these variables. I understand this is exactly what @agjohnson was suggested and wanted me to use now 😄 

Example with a small GIF:

- by default the `:root` selector has no preference over the pre-defined values that Read the Docs defines on `html`
- changing it to `readthedocs-flyout` makes those `--readthedocs-` variables to take effect 🎉 

Peek 2025-01-09 11-28

ericholscher:
> by default the :root selector has no preference over the pre-defined values that Read the Docs defines on html

I thought we wanted folks to be able to define the `:root` selectors?

src/flyout.css
```diff
@@ -1,4 +1,14 @@
-/* New */
+/* Flyout styles */
+
+:host {
+  /* These variables are used more than once, use a local variable so we can set
+   * a default in this addon */
```
ericholscher, line 5: Very useful docstring 💯

src/flyout.js
ericholscher:
```suggestion
```
Seems this is still an issue?

ericholscher (approved):

agjohnson:
> > by default the :root selector has no preference over the pre-defined values that Read the Docs defines on html
> 
> I thought we wanted folks to be able to define the `:root` selectors?

Because we add rules as `:root[data-...=""]` -- specificity `0 1 1` -- the specificity of user selectors will need to exceed this or need to target a child element. The following selectors can override our tool detection CSS rules:

- `html > body`/`html > *` targets a child element of `html`/`:root`
- `readthedocs-flyout` targets a deep child element of `html`/`:root`
- `:root:not(.a)` is an example that has higher specificity (`0 2 0`)

ericholscher:
> > > by default the :root selector has no preference over the pre-defined values that Read the Docs defines on html
> > 
> > 
> > I thought we wanted folks to be able to define the `:root` selectors?
> 
Because we add rules as `:root[data-...=""]` -- specificity `0 1 1` -- the specificity of user selectors will need to exceed this or need to target a child element. The following selectors can override our tool detection CSS rules:

* `html > body`/`html > *` targets a child element of `html`/`:root`
* `readthedocs-flyout` targets a deep child element of `html`/`:root`
* `:root:not(.a)` is an example that has higher specificity (`0 2 0`)

This seems like a great thing to add to the docs 💯

ericholscher:
Added here: https://github.com/readthedocs/readthedocs.org/pull/11888/commits/a371da05a282597ea482e4755e1c0daebefd8f26

agjohnson:
Actually, it's not needed to target `:root`. Using `@layer` we can bump down our defaults in precedence below user contributed selectors, allowing users to just target `:root`. I'm playing around with whether this is useful to drop the `--addons` prefix variables too

agjohnson:
Updated again at https://github.com/readthedocs/addons/pull/487

humitos:
I'm happy we arrived at a great place here! I'm merging this PR so we deploy it today 🚀

PR #474, opened 2024-12-16
API: save a "flag" on `window` object

humitos:
We need to find a way for people subscribing to the Read the Docs data ready event _after_ it was triggered. This is an idea about how to solve this use case.

It adds a "flag" in the `window` object that users can check _before_ subscribing to that event to make usage of the data if the event was triggered already.

* Reference: https://github.com/readthedocs/readthedocs.org/issues/10648#issuecomment-2542003070
* Example: https://github.com/readthedocs/test-builds/blob/full-feature/docs/static/readthedocs.js


src/readthedocs-config.js
```diff
@@ -114,6 +114,10 @@ export function getReadTheDocsConfig(sendUrlParam) {
           // Expose `dataUser` if available or the `data` already requested.
           const dataEvent = dataUser !== undefined ? dataUser : data;
 
+          // Add the data to the window so scripts loaded after the initial
+          // event was fired can still get access to the data
+          window.ReadTheDocsEventData = new ReadTheDocsEventData(dataEvent);
```
agjohnson, line 119: Avoid using `window` as this isn't usable in tests/Node. Use `globalThis` to always reference the correct global.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#description

agjohnson (approved):
Seems good for now, but I think the context API probably will solve this is a cleaner way later.

humitos:
@DimedS We are still experimenting with a different, and more generic approach. I want to continue testing that, before moving forward here since it may replace the idea from this branch. I will let you know when I have some progress on this.

DimedS:
> @DimedS We are still experimenting with a different, and more generic approach. I want to continue testing that, before moving forward here since it may replace the idea from this branch. I will let you know when I have some progress on this.

Thanks, @humitos! Appreciate the update and your efforts!

humitos:
This is an example of how we can use this feature on projects:

```javascript
if (document.querySelector("meta[name='readthedocs-addons-api-version']") === null) {
  // Add a META tag to specify the API version we are expecting
  const meta = document.createElement("meta");
  meta.name = "readthedocs-addons-api-version";
  meta.content = "1";
  document.head.append(meta);
}

// Show a message in the console immediatelly loading this script
console.log("This message comes from a file injected by CustomScript addon.");

function _handleReadTheDocsData (data) {
  console.log("_handleReadTheDocsData function called.");
  console.log("Project slug:", data.projects.current.slug);
}

// The event "readthedocs-addons-data-ready" has been already fired when this script is run.
// We need to check for `window.ReadTheDocsEventData` first, and if it's available
// use that data to call the handler.
if (window.ReadTheDocsEventData !== undefined) {
  _handleReadTheDocsData(window.ReadTheDocsEventData.data());
}

// After that, we subscribe to the Read the Docs Addons event to access data
// on future dispatchs (e.g. when a URL changes on a SPA)
document.addEventListener("readthedocs-addons-data-ready", function (event) {
  _handleReadTheDocsData(event.detail.data());
});

```

Live example: https://test-builds.readthedocs.io/en/full-feature/

PR #475, opened 2024-12-18
EthicalAd: specific placement for Antora and mdBook

humitos:
Screenshot_2024-12-18_16-55-02

Based on https://github.com/readthedocs/addons/pull/473


humitos:
I also added the placement for mdBook as well:

Screenshot_2024-12-18_16-59-00

src/ethicalads.js
```diff
@@ -147,6 +147,28 @@ export class EthicalAdsAddon extends AddonBase {
...
+          placement.setAttribute("data-ea-type", "readthedocs-sidebar");
+          placement.setAttribute("data-ea-style", "image");
+          knownPlacementFound = true;
+        }
+      } else if (docTool.isAntora()) {
+        selector = "aside nav.nav-menu";
+        element = document.querySelector(selector);
+
+        if (this.elementAboveTheFold(element)) {
+          placement.classList.add("ethical-alabaster");
```
ericholscher, line 159: Should we rename this something more generic?

ericholscher (approved):

src/ethicalads.js
humitos: Yes. We should use something like `ethical-light-theme`; considering that we are using it for that use case. However, it's not a 1:1 match. We will need to keep `-alabaster` as well since there are some thing very specific to it, I think. I opened https://github.com/readthedocs/addons/issues/477

ericholscher:
Would be good to merge this :+1:

humitos:
Yeah, thanks for the review. This PR is based on https://github.com/readthedocs/addons/pull/473. We can merge it once we have the underlying one merged as well. The referencing PR seems to be pretty close if we agree on the CSS variables.

PR #476, opened 2024-12-18
Don't wait for DOM ready

humitos:
I understand we don't need for the DOM to be ready to start performing the API requests and initializing our addons.

This PR should improve the load speed a little since it performs the API requests as soon as possible.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,7 +11,6 @@ import * as filetreediff from "./filetreediff";
 import * as customscript from "./customscript";
 import { default as objectPath } from "object-path";
 import {
-  domReady,
   isEmbedded,
   IS_PRODUCTION,
   setupLogging,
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -38,20 +37,17 @@ export function setup() {
...
+    for (const addon of addons) {
+      if (addon.requiresUrlParam()) {
+        sendUrlParam = true;
+        break;
+      }
+    }
 
-        return getReadTheDocsConfig(sendUrlParam);
-      })
+    getReadTheDocsConfig(sendUrlParam)
```
agjohnson, line 50: So, you don't need to wait for the DOM to be ready to _start_ the API request, but we do need the DOM to be ready to do things like detect the doc tool and any other operation that queries the DOM.

I might be missing something here, but it seems like this should add `domReady` into the promise chain _after_ the API request. Does that seem correct?

The effect without this is that there could be network speed, timeouts, etc that allow our API requests to return before the DOM is ready, and then we'd probably get some nondeterministic behavior around DOM queries.

humitos: Yeah, that's true 👍🏼

humitos:
I'm closing this since we are performing some heuristic in the DOM to know what type of documentation tool users are using. It's safer to keep running our code when DOM's ready.

PR #480, opened 2025-01-06
Don't show the `latest` notification if it's the default version

ericholscher:
I think this is maybe a better default than https://github.com/readthedocs/readthedocs.org/pull/11874.

This should at least stop the "getting a notification when I hit the default version" issue. Though maybe it's better to default it to off, instead of this? I think we need to do _something_ to fix the current behavior because it's properly annoying (I hit the latest version warning like 3 times yesterday trying to do basic work, and it was never useful..)

### DIFF
--- a/src/notification.js
+++ b/src/notification.js
@@ -242,8 +242,9 @@ export class NotificationElement extends LitElement {
     //  - if the user is reading the "latest" version: shows a notification to warn
     //    the user about reading the latest development version.
     //  - if the user is reading a non-"stable" version: shows a notification to warn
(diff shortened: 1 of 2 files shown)


humitos (approved):
I think this solves, or at least reduces, the annoyance of this notification 👍🏼 

IIRC, there was some docstring explaining the logic behind _when_ to show the notification, so I think we should update that docstring to reflect this change in behavior.

humitos:
Found it, https://github.com/readthedocs/addons/blob/12028e3d49daee5382949ecf4d903d8220b7bf05/src/notification.js#L239-L247

PR #482, opened 2025-01-08
Flyout: use theme default for `position`

### LINKED ISSUE #434: Implement POC for theme default options
I'm creating a small issue to track the progress of the work we discussed in;

* https://github.com/readthedocs/meta/discussions/161

humitos:
Initial POC to prove if this idea is possible and how we feel about it.

I put together an example of what I understand we want to build here 😅 . In this example, the theme author will make explicitly how they want to use `readthedocs-flyout` HTML tag:

```html
<readthedocs-flyout position="top-right"></readthedocs-flyout>
```

That will add the class `top-right` and show the flyout as follows:

Screenshot_2025-01-08_12-44-20

However, the project author can change the `position` value from the dashboard if they prefer to override the theme's default value:

Peek 2025-01-08 12-46

* Related #51
* Requires https://github.com/readthedocs/ext-theme/pull/554
* Requires https://github.com/readthedocs/readthedocs.org/pull/11891
* Example using https://test-builds.readthedocs.io/en/addons-theme-defaults/
* Closes #434

### DIFF
--- a/src/flyout.js
+++ b/src/flyout.js
@@ -45,6 +45,14 @@ export class FlyoutElement extends LitElement {
       return;
     }
     this.config = config;
+
+    // The "position" is a value that can be defined from the dashboard.
+    // There are two main options: "Default" or a specific value.
+    // When "Default" is used, the value will be grabbed from the HTML element (e.g. explicitly set by the theme author).
+    // In case it's not defined, the value defined in the `constructor` will be used ("bottom-right")
+    this.position =
+      objectPath.get(this.config, "addons.flyout.position", null) ||
+      this.position;
   }
 
   _close() {
--- a/src/flyout.js
+++ b/src/flyout.js
@@ -50,9 +50,14 @@ export class FlyoutElement extends LitElement {
     // There are two main options: "Default" or a specific value.
     // When "Default" is used, the value will be grabbed from the HTML element (e.g. explicitly set by the theme author).
     // In case it's not defined, the value defined in the `constructor` will be used ("bottom-right")
-    this.position =
-      objectPath.get(this.config, "addons.flyout.position", null) ||
-      this.position;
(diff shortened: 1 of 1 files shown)


src/flyout.js
```diff
@@ -45,6 +45,14 @@ export class FlyoutElement extends LitElement {
...
     }
     this.config = config;
+
+    // The "position" is a value that can be defined from the dashboard.
+    // There are two main options: "Default" or a specific value.
+    // When "Default" is used, the value will be grabbed from the HTML element (e.g. explicitly set by the theme author).
+    // In case it's not defined, the value defined in the `constructor` will be used ("bottom-right")
+    this.position =
+      objectPath.get(this.config, "addons.flyout.position", null) ||
+      this.position;
```
ericholscher, line 55: Seems clearer to set this in an `if`, instead of resetting the value to itself?

ericholscher:
```suggestion
    position = objectPath.get(this.config, "addons.flyout.position", null);
    if position:
        this.position = position;
```

humitos: My reading is "get the value from the API or keep the actual one".

I don't have a strong preference, but happy to change it if wanted.

ericholscher: I don't feel super strongly, but think it's more explicit.

humitos:
@agjohnson I would appreciate your review on this PR so we can decide whether or not move forward with this implementation.

src/flyout.js
agjohnson: `position` is a reactive property, so it's best to avoid triggering unneeded updates to the property. Wrapping this with a conditional is probably best. Otherwise this pattern for setting variable defaults is pretty common though.

agjohnson (approved):

PR #484, opened 2025-01-08
Show example of setting attributes and variables at parent DOM

agjohnson:
This moves the inner CSS rules from inside the shadow DOM to the parent
DOM. This allows users to still set a `--readthedocs-font-size` and
similar variables.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,12 +11,15 @@ import * as filetreediff from "./filetreediff";
 import * as customscript from "./customscript";
 import { default as objectPath } from "object-path";
 import {
+  docTool,
   domReady,
   isEmbedded,
   IS_PRODUCTION,
   setupLogging,
   getMetadataValue,
 } from "./utils";
+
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -50,6 +53,28 @@ export function setup() {
...
+              docTool.documentationTool,
+            );
+          }
+          if (docTool.documentationTheme) {
+            elementHtml.setAttribute(
+              "data-readthedocs-tool-theme",
+              docTool.documentationTheme,
+            );
+          }
+        }
```
humitos, line 76: This is where I had the most doubts about and I wasn't able to understand what you were saying. Reading this _in code_ is a lot easier to understand 😄 . Definitely, I read JavaScript better than I listen to English 🙃

humitos (approved):
Thanks a lot for working on this example. It's a lot more clear to me now 👍🏼 

I will merge this PR into mine and continue working and adapting my own PR. It should be pretty close to merge as well.

PR #486, opened 2025-01-08
Fix flex grid and usage of static lengths

### LINKED ISSUE #400: Flyout: use CSS grid to deal with flyout space
In https://github.com/readthedocs/addons/pull/360 we mentioned we may need to use CSS grid or a similar technique to deal with spacing. Mainly when the flyout is collapsed and the version name is long.

One approach could be to use CSS grid areas: https://ishadeed.com/article/css-grid-area/

We need to talk more about this and explore alternatives (like flexbox, https://github.com/readthedocs/addons/pull/360#issuecomment-2289755353). Also note that most of this will change when working on the next iteration of the flyout as a "toolbar" or similar: https://github.com/readthedocs/addons/issues/53

### LINKED ISSUE #485: Flyout: element alignment is wrong with different font sizes
Because the flyout CSS is using a mix of fixed width (`5px` instead of `0.25em`), increasing/decreasing the font size doesn't scale all elements or their positioning/spacing correctly.

Here is a flex grid view of the elements:

image

Notice that the logo doesn't scale and the icons aren't centered.

agjohnson:
- Use flex grid grow/shrink/basis to control the grow/shrink behavior of
  flyout header elements
- Don't wrap header flex items
- Truncate header flex items using ellipsis overflow
- Control height through font height and tame usage of line-height to be
  consistent
- Move padding from header to header>span so that logo doesn't have to
  break out of content height or line height to display larger than the
  font size
- Increase the flyout width and make width controllable by variable

----

- Fixes #485
- Fixes #400
- Refs #394

----

image
image

image
Note here that the logo scales with the font size correctly.

image
image
image


src/flyout.css
```diff
@@ -2,7 +2,7 @@
 
 .container {
   position: fixed;
-  max-width: 300px;
+  max-width: var(--readthedocs-flyout-max-width, 30em);
```
humitos, line 5: It seems this new default makes the flyout widther that it was by default. I don't know if that's an issue, but at least it's a difference to how it was.

humitos: New behavior:

Screenshot_2025-01-09_10-43-44

Old behavior:

Screenshot_2025-01-09_10-43-29

src/flyout.css
```diff
@@ -55,42 +56,51 @@
 
 header {
   display: flex;
-  flex-flow: row wrap;
+  flex-flow: row nowrap;
+  gap: 1em;
```
humitos, line 60: I understand this `gap` is the space between the logo and the current version, right? I noticed that now the current version is aligned to the left when it was aligned to the right before (the screenshot from the previous comment also shows this)

We should probably use `calc()` here if we want to keep the previous behavior. It looks better IMO.

humitos (approved):
This is great! There are specific details that are out of my CSS knowledge, but I tested it locally and it looks great!

I left some comments with the differences I found compared to the current version.

src/flyout.css
agjohnson: Actually, I was only expecting this to grow if the content dictated a wider container. I didn't test with many versions like this. It can be reduced to `25em`, which is about the ratio it's using most of the time now.

src/flyout.css
agjohnson: I wouldn't use calc, no. Flex box has a complex calculation on space and positioning already, there isn't really a way to use calc with flex shrink/grow/basis.

It's possible to just position the item to the right.

agjohnson: Fixed in ae2989c

agjohnson:
Width exaggerated to display flex box spacing/alignment:

image

humitos:
This is great! Thanks for considering my feedback 👍🏼

PR #487, opened 2025-01-09
More variable fixes

agjohnson:
- Wrap doc tool rules with an at-rule `layer`. Layers have less precedence than user CSS styles: https://developer.mozilla.org/en-US/docs/Web/CSS/@layer#description
- Don't disable CSS stylesheet injection, fix import instead
- Avoid exception on missing document attributes, us condition undefined operator

With the `@layer`, user CSS can use `:root { --readthedocs-flyout-font-size: 20px; }` to override the defaults. There is no need to use specificity or inheritance order to override default variables.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,5 @@
+import { CSSResult } from "lit";
+
 import { getReadTheDocsConfig } from "./readthedocs-config";
 import * as notification from "./notification";
 import * as analytics from "./analytics";
@@ -57,7 +59,11 @@ export function setup() {
         const elementHtml = document.querySelector("html");
         if (elementHtml) {
           // Inject styles at the parent DOM to set variables at :root
-          document.adoptedStyleSheets = [doctoolsStyleSheet];
+          let styleSheet = doctoolsStyleSheet;
+          if (doctoolsStyleSheet instanceof CSSResult) {
+            styleSheet = doctoolsStyleSheet.styleSheet;
+          }
+          document.adoptedStyleSheets = [styleSheet];
 
           // If we detect a documentation tool, set attributes on :root to allow
           // for CSS selectors to utilize these values.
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -58,12 +59,11 @@ export function setup() {
         const elementHtml = document.querySelector("html");
         if (elementHtml) {
           // Inject styles at the parent DOM to set variables at :root
-          //
-          // NOTE: We can't include this on testing due to this error:
-          // Failed to set the 'adoptedStyleSheets' property on 'Document': Failed to convert value to 'CSSStyleSheet'.
-          if (!IS_TESTING) {
```
ericholscher, line 64: I think we have this pattern in a couple other places that might be worth removing as well 💯

ericholscher (approved):

humitos (approved):
This works great for me. I tested it using

```
:root {
  --readthedocs-flyout-font-size: 2rem;
}
```

and I can see the changes immediately.

I proposed an approach to use `<addons>.defaults.css` in #488 that we can include here (or I can do it in another PR) to have these CSS following a standardized pattern.

src/index.js
agjohnson: I don't think we need to change this pattern elsewhere yet, just fixing the import for tests rather than masking this code from tests.

The error comes from web-test-runner using Rollup and our builds using Webpack. Both these handle the `import { styleSheet } from "flyout.css"` by exporting differing classed objects.

agjohnson:
I wouldn't use a separate, secondary CSS files for variable defaults for each addon unless we have to. I might even just prefer one CSS file if we're doing that. Here's an example of what it looks like if we author CSS variable defaults in each addon CSS file:

- https://github.com/readthedocs/addons/pull/492/commits/5d2a33f056aae4a019383dfe312e7291e57d4838

This is just an example of what we can work towards, I don't think we need to hold up this PR or the underlying to continue working on this pattern.

humitos:
Cool, I'll merge this PR into mine and try to wrap mine as well so we can move forward.

Issue #488, opened 2025-01-10
CSS: cleaner pattern for variable use


agjohnson:
Here are a couple options for cleaning up our CSS variable patterns. I started to experiment with some of these to see how it felt to override them and author rules around them. I wanted to note this before moving on.

## Current issues

- We redefine the defaults used for variables in more than one place -- we use `font-family: var(--readthedocs-font-family, "Lato", ...)` in multiple places.
- We use `--addons` prefixed variables inconsistently, so the pattern feels unclear but also redundant.
- We create more variables than necessary, for instance `--readthedocs-flyout-dd-font-size`.
- Usage of `calc()` instead of `em` rules makes variable values probably more confusing than avoiding `em` helps. `em` feels safer in a shadow DOM, so `rem` doesn't feel as necessary as working in the parent DOM.
- We haven't decided or dictated what selectors we want users to use to override variables.

### Set variable defaults on parent DOM `:root`

This would be a stylesheet added to `document.adoptedStyleSheets`:

```css
@layer defaults {
  :root {
    --readthedocs-font-size: 16px;
    --readthedocs-font-family: monospace;
    --readthedocs-flyout-font-size: var(--readthedocs-font-size);
  }
}
```

And when we use the variable, we don't need to set a default value now:

```css
:host {
  font-size: var(--readthedocs-flyout-font-size);
}
```

And users can override these variables with:

```css
:root {
  --readthedocs-flyout-font-size: 24px;
}
```

- **Good**: We only define the variable defaults once
- **Good**: No intermediate `--addons` variables
- **Meh**: We have to defince the default values outside each addon

### Users set defaults on our elements, not :root

Another approach would be to set the variable defaults on each addon Element `:host`:

```css
/* flyout.css */
:host {
  --readthedocs-flyout-font-size: var(--readthedocs-font-size);
  --readthedocs-flyout-color: red;
}

.container {
  font-size: var(--readthedocs-flyout-font-size);
}
```

This would differ from the approach above in that users would override variables on elements, not on the parent DOM `:root`:

```css
readthedocs-flyout {
  --readthedocs-flyout-font-size: 24px;
}
```

This works because the parent DOM `readthedocs-flyout` is the same element as the shadow DOM `:host`. We can't do the above pattern if we instruct users to override variables at `:root`, this is the reason we have the `--addons` prefixed variables.

- **Good**: Variable defaults are set in each addon and right at the top of the CSS. Super easy to work with.
[...]

PR #492, opened 2025-01-13
Example to show settings all varialbes at :root explicitly

agjohnson:
Refs #488, this is the first option. Instead of setting each variable at the
point in time we use it (including lots of duplicates) we set _all_ variables at the
:root in the parent DOM. The drops the need for the intermediate `--addons-...`
variables and gives a point where these variables can be default in one place.

The downside to this generally is that every variable needs to be defined
outside each addon CSS file. We can avoid this and define variables inside the
addon CSS by extracting what we need from the addon CSS.

There are a few ways to do this too though:

- Export a `variableDefaults` using `css`/`CSSStyleSheet` from `flyout.js`. This
  still separates variable definition from CSS variable use.
- A third CSS file for every addon. Similar to just having a `defaults.css` but
  we'd have 8 or so of these files instead of one.


humitos:
Would you mind rebasing this PR since it got some conflicts with the last merges?

agjohnson:
Yeah I can try rebasing this this week. Does this all make sense so far do you think?

agjohnson:
This was a bit difficult to rebase, non of the conflicts were mine. I pushed up a separate PR so I didn't ruin this one:

- #545

PR #494, opened 2025-01-14
Visual Diff UI v1

ericholscher:
This adds a v1 Visual Diff UI.
It does the following:

* Adds a UI element to track Docdiff state, and exposes filetreediff as a dropdown
* Updates the URL to track diff state
* Updates the UI element to show state based on events (eg. hotkeys)
* Shows the current file in the dropdown if it's open, otherwise a generic message

Fixes https://github.com/readthedocs/meta/issues/171

## Demo

Screenshot 2025-01-14 at 12 28 49 PM

https://github.com/user-attachments/assets/768c7bbf-e8f2-4753-8d69-1ac718b52b69

### DIFF
--- a/src/docdiff.js
+++ b/src/docdiff.js
@@ -198,6 +198,11 @@ export class DocDiffElement extends LitElement {
       return null;
     }
 
+    // Update URL to include the diff parameter
+    const url = new URL(window.location.href);
+    url.searchParams.set(DOCDIFF_URL_PARAM, "true");
+    window.history.replaceState({}, "", url);
+
     this.enabled = true;
     this.originalBody = document.querySelector(this.rootSelector);
     return this.compare();
@@ -208,6 +213,11 @@ export class DocDiffElement extends LitElement {
       console.debug("Ignoring disableDocDiff: it was already disabled");
       return null;
     }
+
+    // Remove diff parameter from URL
+    const url = new URL(window.location.href);
+    url.searchParams.delete(DOCDIFF_URL_PARAM);
+    window.history.replaceState({}, "", url);
 
     this.enabled = false;
     document.querySelector(this.rootSelector).replaceWith(this.originalBody);
@@ -238,6 +248,7 @@ export class DocDiffElement extends LitElement {
       this._handleHideDocDiff,
     );
   }
+
   disconnectedCallback() {
     document.removeEventListener(
       EVENT_READTHEDOCS_DOCDIFF_ADDED_REMOVED_SHOW,
--- a/src/filetreediff.css
+++ b/src/filetreediff.css
@@ -1,86 +1,39 @@
-:host > div {
-  margin: 1rem 0rem;
-  padding-top: 1rem;
-  padding-bottom: 1rem;
-  overflow: auto;
-  border-radius: 0.5rem;
-  font-family: "Lato", "proxima-nova", "Helvetica Neue", "Arial", "sans-serif";
-  font-size: 1rem;
-  color: rgb(64, 64, 64);
-  background-color: rgb(234, 234, 234);
-}
-
-:host > div > div.content > ul {
-  margin: 0;
-  padding-left: 20px;
-}
-
-:host(.toast) > div {
+.file-dropdown {
   position: fixed;
-  padding-top: calc(1rem * 0.75);
-  padding-bottom: calc(1rem * 0.75);
-  margin: 0.75rem 0rem;
-  top: 7rem;
-  right: 2rem;
-  z-index: 1750;
-  font-size: calc(1rem * 0.85);
-  width: 35rem;
-  max-width: calc(100vw - 4rem);
-}
-
-@media (max-width: 640px) {
-  :host(.toast) > div {
-    right: 0.5rem;
-  }
-}
-
-:host > div > svg.header.icon {
-  height: calc(1rem * 2);
-  padding: 0.5rem 1.5rem;
-  float: left;
-}
-
-:host(.toast) > div > svg.header.icon {
-  height: calc(1rem * 1.5);
-}
-
-:host > div a {
-  color: rgb(8, 140, 219);
-  text-decoration: none;
-}
-
-:host > div > .title {
-  padding: 0.25rem 1rem;
-  margin-bottom: 0.25rem;
-  line-height: 1rem;
-  font-weight: bold;
+  top: 0;
+  right: 1rem;
+  background-color: #f8f9fa;
+  padding: 0.25rem 0.75rem;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+  z-index: 2000;
+  border-radius: 0 0 6px 6px;
   font-size: 0.85rem;
 }
-:host > div > div.content {
-  line-height: 1rem;
-  font-size: calc(1rem * 0.85);
-  padding-left: 66px;
+
+.diff-controls {
+  display: flex;
+  gap: 0.75rem;
+  align-items: center;
 }
 
-:host(.toast) > div > .title {
-  padding: 0rem 1rem;
+.diff-controls select {
+  flex: 1;
+  padding: 0.25rem 0.5rem;
+  border: 1px solid #ddd;
+  border-radius: 4px;
+  appearance: revert;
 }
 
-:host > div > .title > .right {
-  float: right;
-}
-
-:host > div > .title > .right > svg {
-  display: inline-block;
-  height: 1rem;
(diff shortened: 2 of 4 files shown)


src/filetreediff.js
```diff
@@ -13,103 +15,134 @@ export class FileTreeDiffElement extends LitElement {
...
+          <label class="diff-checkbox">
+            <input
+              type="checkbox"
+              .checked=${this.docDiffEnabled}
+              @change=${this.handleToggleDiff}
+            />
+            Show diff
+          </label>
+          <select id="file-select" @change=${this.handleFileChange}>
+            <option value="" ?selected=${!hasCurrentFile}>Files changed</option>
```
humitos, line 107: I think it's not possible, but I'm asking anyways. I found that `Files changed` is an option the user can select but does nothing. Would it be possible to make it non-selectable?

To reproduce:

1. open a PR
2. click on Files changed
3. select a file from the dropdown
4. a new page loads
5. click on the files dropdown 
6. click on Files changed
7. ... nothing happens

Does it make sense to _remove_ the "Files changed" title when a file is selected?

src/index.js
```diff
@@ -29,12 +29,13 @@ export function setup() {
...
     // listen to to update its internal state.
+    // Same for FileTreeDiff.
     hotkeys.HotKeysAddon,
-    docdiff.DocDiffAddon,
+    filetreediff.FileTreeDiffAddon,
 
     linkpreviews.LinkPreviewsAddon,
-    filetreediff.FileTreeDiffAddon,
     customscript.CustomScriptAddon,
+    docdiff.DocDiffAddon,
```
humitos, line 38: I'm not sure to understand this change. It seems that hotkeys addon was already initialized before filetreediff. Is there any other reason why you needed to re-order them here?

src/filetreediff.css
```diff
@@ -1,86 +1,39 @@
...
-  margin-bottom: 0.25rem;
-  line-height: 1rem;
-  font-weight: bold;
+  top: 0;
+  right: 1rem;
+  background-color: #f8f9fa;
+  padding: 0.25rem 0.75rem;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+  z-index: 2000;
+  border-radius: 0 0 6px 6px;
```
humitos, line 9: @agjohnson what would be the pattern here to add `--readthedocs-*` CSS variables? what are the _correct/required_ ones to add? All of those with `*color*` on its name? Should all the "spacing" values be calculated from `--readthedocs-filetreediff-font-size`?

src/filetreediff.css
```diff
@@ -1,86 +1,39 @@
...
-  background-color: rgb(234, 234, 234);
-}
-
-:host > div > div.content > ul {
-  margin: 0;
-  padding-left: 20px;
-}
-
-:host(.toast) > div {
+.file-dropdown {
```
humitos, line 1: I understand we are not using CSS classes for this, but more strict selectors instead. Example: `:host > div > div > label` and similars. I'm not 100% about the benefits, but just mentioning what's the pattern we have been following for other addons -- cc @agjohnson

humitos (commented):
This is, without any doubt, a lot better than what we had 😄 . It's tiny, simple, and usable.

src/index.js
ericholscher: I think it's actually docdiff that needs to run last, because it's triggering the event based on the URL param and filetreediff is now listening to that to keep state. It's messy keeping state in multiple places with events like this, and I'm not convinced it's a good long term solution vs. having the docdiff addon keep the "enabled/disabled" state that other addons can query.

src/filetreediff.css
ericholscher: Yea, I didn't have any understanding of the current pattern, so just went with something that worked. If we have something explaining the approach we're taking, happy to follow them.

src/filetreediff.js
ericholscher: Yea, we can definitely do both I think.

src/index.js
ericholscher: https://github.com/readthedocs/addons/blob/47645b013724cdf244716b549a5baa28409fafcb/src/docdiff.js#L105-L111

src/filetreediff.js
```diff
@@ -13,103 +15,136 @@ export class FileTreeDiffElement extends LitElement {
...
+            <input
+              type="checkbox"
+              .checked=${this.docDiffEnabled}
+              @change=${this.handleToggleDiff}
+            />
+            Show diff
+          </label>
+          <select id="file-select" @change=${this.handleFileChange}>
+            <option value="" ?selected=${!hasCurrentFile} disabled>
+              Files changed:
```
humitos, line 108:
```suggestion
              Files changed
```

IMO, it looks nicer without the colon.

src/docdiff.js
```diff
@@ -198,6 +198,11 @@ export class DocDiffElement extends LitElement {
       return null;
     }
 
+    // Update URL to include the diff parameter
+    const url = new URL(window.location.href);
+    url.searchParams.set(DOCDIFF_URL_PARAM, "true");
+    window.history.replaceState({}, "", url);
```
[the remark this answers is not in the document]
humitos: Maybe we need to check if the `path` part of the URL changed, without considering the query part in https://github.com/readthedocs/addons/pull/455/files#diff-3274f1a37032fb0ae4e2823def0007c634e869ae0dfc304ff6a12c36513c3a52R185

That should probably be enough 👍🏼

src/index.js
humitos: > docdiff addon keep the "enabled/disabled" state that other addons can query.

This will have a similar problems. If the state changes on docdiff, you need to communicate this somehow to the other addons to update their UI as well. I don't think you can have a "reactive property" based on _other web-component property_.

ericholscher:
@humitos if you're happy with the overall concept here, I think we 👍 it and ship it, and we can iterate in dev. I'd like to start using visual diff more on WTD, and this UI makes it not a blocker :)

src/docdiff.js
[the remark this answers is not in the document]
humitos: #455 was merged already. We should open an issue for this so we don't loose. I can probably take a look at it next week, it shouldn't be too much work.

humitos (approved):
I re-review this and I don't seeing anything blocking us to deploy this work. It's better to what we currently have and it has to be enabled per-project -- which looks safe.

Once it's deployed, we can ping some users so they can start testing it and give us some feedback (e.g. CPython folks are waiting for this, actually).

Let's ship it and iterate over it with the feedback we receive and our own as well 👍🏼

src/docdiff.js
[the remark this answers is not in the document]
ericholscher: https://github.com/readthedocs/addons/issues/506

ericholscher:
Excited to test this out more! Always happy to update the code once we have patterns formalized.

src/filetreediff.css
agjohnson: Whatever we're doing in other addons right now is fine. There isn't any issue with adding CSS variables that we'll change later especially if we aren't documenting them.

src/filetreediff.js
```diff
@@ -13,103 +15,136 @@ export class FileTreeDiffElement extends LitElement {
...
+        <div class="diff-controls">
+          <label class="diff-checkbox">
+            <input
+              type="checkbox"
+              .checked=${this.docDiffEnabled}
+              @change=${this.handleToggleDiff}
+            />
+            Show diff
+          </label>
+          <select id="file-select" @change=${this.handleFileChange}>
```
agjohnson, line 106: `file-select` seems unused, are we doing something with this?

If this needs to be targeted with an event, use Lit `@[event]` attributes instead. It doesn't look like this is the case though.

src/filetreediff.css
```diff
@@ -1,86 +1,39 @@
...
+  align-items: center;
 }
 
-:host > div > .title > .right {
-  float: right;
+.diff-controls select {
+  flex: 1;
+  padding: 0.25rem 0.5rem;
+  border: 1px solid #ddd;
+  border-radius: 4px;
```
agjohnson, line 23: Generally, all units should be `em`/`rem` to avoid breaking scaling of these elements with font size -- `px` won't scale up if users override the font-size larger.

agjohnson (commented):
Just noting some general patterns to follow, otherwise this PR looks great. I'm glad to have the update to replace the notification UI.

PR #499, opened 2025-01-15
Use ContextRoot at document.html instead

agjohnson:
Additional elements shouldn't be needed in our structure. This is how we can use ContextRoot with `document.html` as the context root for providers/consumers of the config context.

Both the flyout and the notifications load for me with this change.


humitos (approved):
Thanks for putting an example for this. I will merge this PR into mine now and continue from there.

PR #500, opened 2025-01-16
Flyout: define `position="inline"`

humitos:
Allow theme developers to define `position="inline"` to adapt the flyout to its container. It uses `bottom: 0` to make it open up side, and it should probably use `top: 0` to make it open down side.

### DIFF
--- a/src/flyout.css
+++ b/src/flyout.css
@@ -40,6 +40,20 @@
   top: 50px;
 }
 
+.container.inline {
+  width: 100%;
+  padding: 0;
+  margin: 0;
+  position: absolute;
+  bottom: 0;
+}
+
+@media screen and (max-width: 768px) {
+  .container.inline {
+    max-width: 100%;
+  }
+}
+
 :host > div {
   font-family: var(
     --readthedocs-flyout-font-family,
--- a/src/flyout.js
+++ b/src/flyout.js
@@ -26,7 +26,6 @@ export class FlyoutElement extends LitElement {
   static properties = {
     config: { state: true },
     opened: { type: Boolean },
-    floating: { type: Boolean },
     position: { type: String },
   };
 
@@ -37,7 +36,6 @@ export class FlyoutElement extends LitElement {
 
     this.config = null;
     this.opened = false;
-    this.floating = true;
     this.position = "bottom-right";
     this.readthedocsLogo = READTHEDOCS_LOGO;
   }
@@ -327,7 +325,7 @@ export class FlyoutElement extends LitElement {
   }
 
   updateCSSClasses() {
-    this.classes = { floating: this.floating, container: true };
+    this.classes = { container: true };
     this.classes[this.position] = true;
   }
 


src/flyout.js
```diff
@@ -26,7 +26,6 @@ export class FlyoutElement extends LitElement {
   static properties = {
     config: { state: true },
     opened: { type: Boolean },
-    floating: { type: Boolean },
```
ericholscher, line 29: Was this not used?

ericholscher (approved):
Simple enough!

src/flyout.js
humitos: Nope

src/flyout.css
```diff
@@ -40,6 +40,20 @@
   top: 50px;
 }
 
+.container.inline {
+  width: calc(100% - 20px);
```
agjohnson, line 44: Any time we're working with width, padding, margins, or positioning, it's best to use relative units -- `em`/`rem` -- instead of `px`. Basically always avoid `px` unless we truly want pixel units -- `border: 1px solid red`.

src/flyout.css
```diff
@@ -40,6 +40,20 @@
...
+  padding: 0 10px;
+  margin: 0;
+  position: absolute;
```
agjohnson, line 47: Hrm, so this isn't an inline element then, it's still going to be fixed to the bottom of the screen. Inline would be `display: block-inline;` and the user can put this element anywhere in their DOM.

If the user wanted to put this element anywhere but the bottom left corner, this change doesn't allow that.

This is more like `bottom left attached` than `inline`.

agjohnson: Also, I'm not quite sure on `absolute` here, this is different from `fixed` and will give some fun effects -- this is just on a page with long content:

https://github.com/user-attachments/assets/b85223a3-9399-4962-a3ac-0e6ad69bf587

What was the need for the change here if you are aiming for a fixed position element?

agjohnson:
Here is a very rough example of inline positioning: https://github.com/readthedocs/addons/pull/501

It doesn't take much more but there is some cleanup required and working into themes might require experimenting with overflow clipping on the theme elements.

src/flyout.css
humitos: The flyout behaves as an absolute position based on a relative position from the parent element -- that's why it scrolls when you scrolls the page maintaining its position. It's not a floating fixed.

humitos:
This is an example of what I currently have. It shows the _current default floating bottom right_ position, then the _bottom inline position_ and finally the _top inline position_.

I think we should probably change `bottom` and `top` by `down` and `up` classes, meaning what's the direction the flyout opens.

Peek 2025-01-20 14-21

humitos:
### Read the Docs theme

Peek 2025-01-20 15-20

humitos:
### Sphinx with Furo theme

Peek 2025-01-20 16-35


* Furo fork that injects the flyout: https://github.com/pradyunsg/furo/compare/main...humitos:furo:readthedocs-flyout-inline?expand=1

ericholscher:
The furo theme isn't showing properly in the sidebar, right?

humitos:
It works for me. I uploaded a GIF in my previous comment. Why are you asking?

ericholscher:
It's overlapping?

image

humitos:
Ah, yes. I think that is responsibility of the theme author, not the flyout itself -- but I'm not 100% sure. I understand the flyout container has to be adjusted with CSS properties to make the flyout behave as expected.

humitos:
By the way, removing the `padding: 0 10px` from `.container.inline` makes the flyout to match the sidebar size on Furo. However, it doesn't look as we want because the logo and the caret has no padding and touch the borders. Also, it doesn't cover all the width of the sidebar on Read the Docs theme.

humitos:
I was able to fix the in Furo adding CSS in the flyout container without changing anything in our code. So, this is responsibility of the theme developer.

* Branch on forked Furo repository: https://github.com/pradyunsg/furo/compare/main...humitos:furo:readthedocs-flyout-inline?expand=1

Peek 2025-01-21 10-43

humitos:
I tested this implementation on `insipid` theme as well.

Note that it works fine, but it requires to modify two CSS properties from inside the flyout itself. These properties are calculated dynamically, but I wasn't able to find a formula that works for all these themes. I'm sure there will be a way to calculate this correctly, but I don't know how to do it cc @agjohnson 

These properties are: `.container.up.inline main { bottom: }` and `.container.inline main { margin-left: }`. Here you have an example: https://github.com/mgeier/insipid-sphinx-theme

humitos:
Boom! I was able to find the CSS combination that works 👍🏼 

Example for Insipid, Read the Docs and Furo themes.

Peek 2025-01-21 12-36

src/flyout.css
agjohnson: To clarify more here, in our theme the parent element is `fixed`, which is why the combination of `absolute` effectively fixes the flyout to the bottom of the screen. If users try this styling without that specific combination of positioning, they'll get effects like I showed.

My main point was that it's effectively fixed positioned instead of inline though, I think this is on a better path inline positioned already though.

src/flyout.js
```diff
@@ -327,10 +325,33 @@ export class FlyoutElement extends LitElement {
...
+      }
+
+      if (this.position.includes("down")) {
+        main.style.setProperty(
+          "margin-top",
+          `${header.getBoundingClientRect().height}px`,
+        );
+      }
+    }
+  }
```
agjohnson, line 353: So we should always avoid doing anything with JS, especially with `Element.style`. What this property does is add inline `style=""` attribute to the element, which causes trouble with secure CSP rulesets and causes problems when users try to override our styles.

What value do we not know that we need to perform calculations in JS to get?

humitos: Do you know how to do that with pure CSS? 🤔

agjohnson: In CSS this seems like it should just be addition. If we're mixing units, we should change them all to `em` so that the math is just a simple `line-height + margin-bottom + margin-top + padding-bottom + padding-top`.

Unless there is something I'm missing here?

humitos: > What value do we not know that we need to perform calculations in JS to get?

I need to calculate the `main.margin-bottom`'s property based on the `header.height`'s property. How I can do that in CSS?

`header.height` is not static since it will be based on the container where the theme author inject the `<readthedocs-flyout>` element. How I can get that value with pure CSS?

humitos:
I dug a little more here and I think the different behavior between our theme and furo theme is that furo uses flex for the navigation bar and we don't. This is just a guess because the they it expands and fulfill the spacing.

Note that removing the hack JS code and extra CSS, the flyout grows properly on furo. The header goes to the top as it should be, instead of keeping below the main content as I showed in the other examples.

Peek 2025-01-23 11-30

However, with this code, the flyout in our theme grows to the bottom when clicked and its not shown in the page. I tried playing with flex in our theme, but I wasn't able to get it working 😢

humitos:
Yeap, using `display: flex` and `flex-direction: column` on the nav of our theme made the trick! It expands properly now --leaving the header of the flyout above the main content of it.

Peek 2025-01-23 11-47

src/flyout.js
agjohnson: We don't need to calculate it because we _should know this value_.

> header.height is not static since it will be based on the container where the theme author inject the <readthedocs-flyout> element. 

This doesn't seem correct. If the parent element is 10em, we get this effect if we try to fill vertical space with 100% height:

image

We instead want the header to be static height, and derived from `font-size`. We should be able to determine the value needed for `margin-bottom` just by math based on this.

agjohnson:
Seems you lost the up/down distinction, this should open up:

image

There is some positioning issues on the inline flyout inner element too it seems:

image

src/flyout.js
humitos: Ah, I thought we wanted height and width to be dynamic and based on its container. If we want a fixed height, that should be easier to calculate, yeah.

humitos:
When opening "up", how do we want the flyout to behave?

|Move the header to the top|Keep the header at the bottom|
|----|----|
|Screenshot_2025-01-28_13-01-44|Screenshot_2025-01-28_13-01-24|

If we want "Move the header to the top", I wasn't able to figure it out how to do that when `.inline`.

agjohnson:
Noted on our last call, both options have some benefits I think:

- Keeping the header on top is the position historically. Changing this is maybe unexpected?
- Keeping the header position makes the flyout UX match a standard dropdown -- the dropdown itself doesn't move when you expand the dropdown menu.

It shouldn't be impossible to keep the positioning at the top with some more CSS either though. This is something I can help look at.

humitos:
Cool! I will wait for your help on this. Maybe we can schedule a day for pairing on this: doing an initial review of my code and a final polish together. That way, I can learn your thinking process at the same time.

humitos:
I "created an issue to create issues" for the required work. I'm linking it here, so we have all of these issues backlinked somewhere.

* https://github.com/readthedocs/addons/issues/569

I'm closing this one for now since it doesn't make too much sense to move forward with this approach without further discussion and experiment with the new UI.

PR #503, opened 2025-01-22
Heuristic to detect VitePress

humitos:
- Update `DocumentationTool` class to detect VitePress
- Add better ad positioning for this tool

Screenshot_2025-01-22_10-31-20


ericholscher (approved):
I'd love to have doc quickstarts for all the tools here, should add vitepress to the roadmap, especially if we have a demo user repo already using it fully instead of just a test-builds project.

PR #508, opened 2025-01-23
Add CSS variables

ericholscher:
Branches from https://github.com/readthedocs/addons/pull/507

This is just a basic attempt at adding variables -- wasn't sure what the "right" ones to add would be.


humitos (approved):
I think these are fine for now 👍🏼

PR #511, opened 2025-01-27
Make notification copy a bit cleaner.

ericholscher:
This also makes the latest version warning 1 line, taking less vertical space.

### DIFF
--- a/src/notification.js
+++ b/src/notification.js
@@ -297,7 +297,7 @@ export class NotificationElement extends LitElement {
           This is the latest development version ${this.renderCloseButton()}
         </div>
         <div class="content">
-          Some features may only be avilable in the
+          Some features may not yet be available in the
           <a href="${this.urls.stable}">stable version</a>.
         </div>
       </div>
--- a/src/notification.js
+++ b/src/notification.js
@@ -320,7 +320,7 @@ export class NotificationElement extends LitElement {
         </div>
         <div class="content">
           Refer to the <a href="${this.urls.stable}">stable version</a> for the
-          most up-to-date version.
+          most current documentation.
         </div>
       </div>
     `;
--- a/src/notification.js
+++ b/src/notification.js
@@ -294,14 +294,11 @@ export class NotificationElement extends LitElement {
       <div>
         ${iconFlask.node[0]}
         <div class="title">
-          This is the <span>latest development version</span>
-          ${this.renderCloseButton()}
+          This is the latest development version ${this.renderCloseButton()}
         </div>
         <div class="content">
-          Some features may not yet be available in the published stable
-          version. Read the
-          <a href="${this.urls.stable}">stable version of this documentation</a
-          >.
+          Some features may only be avilable in the
+          <a href="${this.urls.stable}">stable version</a>.
         </div>
       </div>
     `;
@@ -322,10 +319,8 @@ export class NotificationElement extends LitElement {
           ${this.renderCloseButton()}
         </div>
         <div class="content">
-          You may be reading an old version of this documentation. Read the
-          <a href="${this.urls.stable}"
-            >latest stable version of this documentation</a
-          >.
+          Refer to the <a href="${this.urls.stable}">stable version</a> for the
+          most up-to-date version.
         </div>
       </div>
     `;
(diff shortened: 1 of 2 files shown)


src/notification.js
```diff
@@ -294,14 +294,11 @@ export class NotificationElement extends LitElement {
       <div>
         ${iconFlask.node[0]}
         <div class="title">
-          This is the <span>latest development version</span>
-          ${this.renderCloseButton()}
+          This is the latest development version ${this.renderCloseButton()}
```
humitos, line 297: Is there a reason to remove the `span` from here? I think it adds some CSS differentiation.

src/notification.js
```diff
@@ -294,14 +294,11 @@ export class NotificationElement extends LitElement {
...
         </div>
         <div class="content">
-          Some features may not yet be available in the published stable
-          version. Read the
-          <a href="${this.urls.stable}">stable version of this documentation</a
-          >.
+          Some features may only be avilable in the
+          <a href="${this.urls.stable}">stable version</a>.
```
humitos, line 301:
```suggestion
          Some features may not yet be available in the
          <a href="${this.urls.stable}">stable version</a>.
```

It seems you changed the meaning of the phrase.

tests/notification.test.html
```diff
@@ -129,29 +133,14 @@
...
-</div>
-      `);
+            const title = element.shadowRoot.querySelector("div.title");
+            expect(title.innerText).to.be.equal(
+              "This is the latest development version"
+            );
+            const content = element.shadowRoot.querySelector("div.content");
+            expect(content.innerHTML).to.include(
+              '<a href="https://project.readthedocs.io/en/stable/section/page.html">'
+            );
```
humitos, line 143: I think it's important to check the whole HTML we are expecting to show here. If you don't want to write it yourself, you can use a snapshot, as we do for the flyout: https://github.com/readthedocs/addons/blob/3815b3a10999ab854ce4f5fc92ed393db1f558ef/tests/__snapshots__/flyout.test.snap.js

Usage in test: https://github.com/readthedocs/addons/blob/3815b3a10999ab854ce4f5fc92ed393db1f558ef/tests/flyout.test.html#L285

You can create them by running `web-test-runner --update-snapshots` and shouldn't be modified manually.

humitos (commented):
The current copy looks good to me, so I'm -0 on the new copy. The new one is more condensed, which I'm not sure it's _better_.

* There is one message that changes it meaning and I think it's wrong.
* We should keep the check for the full HTML render, as we were doing.

src/notification.js
```diff
@@ -322,10 +319,8 @@ export class NotificationElement extends LitElement {
           ${this.renderCloseButton()}
         </div>
         <div class="content">
-          You may be reading an old version of this documentation. Read the
-          <a href="${this.urls.stable}"
-            >latest stable version of this documentation</a
-          >.
+          Refer to the <a href="${this.urls.stable}">stable version</a> for the
+          most up-to-date version.
```
agjohnson, line 323: Small word change but "most up to date version" seems more like the job of `latest`. With avoiding describing versions as releases, this feels more descriptive:

```suggestion
          Refer to the <a href="${this.urls.stable}">stable version</a> for the
          most current documentation.
```

src/notification.js
ericholscher: I didn't see anything targeting it.

tests/notification.test.html
humitos: We should have at least one test that check the full HTML. It's important to be sure what we are putting in front of our users and that the layout it's not broken.

The easy way to do this is using snapshots as I suggested. It does 99% of the work automatically.

ericholscher: Yea, but if you change the HTML, and then just run a script to update the test based on what is rendered, then any issues will be missed because you're just overwriting with the updated HTML that broke, because any time you change anything you're just bulk rewriting the test?

humitos: If you change the layout, the test will fail. You can check why it fails, and if it fails because of HTML layout, you re-ran the generation of snapshot.

If you didn't change the layout, and the test fails, the test just found a bug 😉

PR #514, opened 2025-01-29
Filetreediff: don't show "Show diff" when docdiff is disabled

### LINKED ISSUE #513: Filetreediff: "Show diff" checkbox shouldn't be shown when DocDiff is disabled
When the user disabled `DocDiff` from the addons configuration, the checkbox "Show diff" in the filetreediff addon should not be shown.

Image

humitos:
Closes #513

### DIFF
--- a/src/filetreediff.js
+++ b/src/filetreediff.js
@@ -63,7 +63,7 @@ export class FileTreeDiffElement extends LitElement {
   }
 
   renderDocDiff() {
-    if (objectPath.get(this.config.doc_diff.enabled, false)) {
+    if (objectPath.get(this.config, "addons.doc_diff.enabled", false)) {
       return html`
         <label>
           <input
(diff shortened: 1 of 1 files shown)


ericholscher (approved):

ericholscher:
I don't really understand what the use case is for disabling visual diff. It seems like having it be able to be enabled and disabled on the page by the user is the best UX. Is there some reason that we're imagining people wanting to turn this off when it isn't enabled by default?

humitos:
I know there was some people disabling it due to hotkeys issues: "pressing `d` on a text box doesn't work" or similar. However, that could be solved by disabling hotkeys addons instead.

Anyways, basically, I think that people will disable addons because different reasons: don't like it, doesn't work as they expect, confuse their users, etc. So, if we allow to disable the addons, we should check if enable/disable and not expose it to them.

PR #519, opened 2025-02-03
Visual diff: jump between diff chunks

### LINKED ISSUE #518: Visual diff: scroll to diff chunks
I found myself wanting the "Show diff" to scroll to the first change when I click it. That way, I can immediately see the changes in the pages. Ideally, I'd like to have extra up/down arrows next to "Show diff" so I can jump between chunks.

Here is an example:

https://github.com/user-attachments/assets/ec3534bb-93b4-4e4c-a393-fa1795d592c4

```javascript
window.scrollTo(0, document.querySelectorAll("p:has(.doc-diff-removed), p:has(.doc-diff-added)").offsetTop - 150)
```

Note that we can't switch between each of the `.doc-diff-removed` and `.doc-diff-added` as-is because there is a bunch of them. It could even be one per word changed. So, we use `p:has(.doc-diff-*)` to detect "chunk of changes" inside the same paragraph.

humitos:
Allow the user to click "up" or "down" arrows to jump between the diff chunks of the page. I did it pretty quickly, so it may need some refactor polish.

[Peek 2025-02-03 16-25.webm](https://github.com/user-attachments/assets/f191c3ef-d601-4143-8259-91ffa8ddffa2)

Closes #518

### DIFF
--- a/src/filetreediff.css
+++ b/src/filetreediff.css
@@ -57,3 +57,9 @@
 :host > div > div > label > input {
   margin: 0;
 }
+
+:host > div > div > span {
+  display: flex;
+  width: 0.5rem;
+  cursor: pointer;
+}
--- a/src/filetreediff.js
+++ b/src/filetreediff.js
@@ -1,3 +1,5 @@
+import { library, icon } from "@fortawesome/fontawesome-svg-core";
+import { faArrowUp, faArrowDown } from "@fortawesome/free-solid-svg-icons";
 import { html, nothing, LitElement } from "lit";
 import { default as objectPath } from "object-path";
 import styleSheet from "./filetreediff.css";
@@ -24,6 +26,17 @@ export class FileTreeDiffElement extends LitElement {
     super();
     this.config = null;
     this.docDiffEnabled = false;
+
+    this.chunkIndex = null;
+    library.add(faArrowDown);
+    library.add(faArrowUp);
+
+    this.iconArrowUp = icon(faArrowUp, {
+      classes: ["icon"],
+    });
+    this.iconArrowDown = icon(faArrowDown, {
+      classes: ["icon"],
+    });
   }
 
   loadConfig(config) {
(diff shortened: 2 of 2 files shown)


stsewd:
An idea for another improvement here: temporarily flash each chunk when navigating, so it is easy to see on which chunk you are jumping on.

ericholscher:
This is what GH does for a similar UI -- I think the counts are useful, as well as highlight that @stsewd mentioned: 

Screenshot 2025-02-03 at 7 38 01 AM

humitos:
Needs a little more work in the scrolling part, but that's the idea:

[Peek 2025-02-03 17-46.webm](https://github.com/user-attachments/assets/1499d64a-4098-4b37-ada3-ee19ad36f373)

src/docdiff.document.css
```diff
@@ -10,3 +10,7 @@
   background-color: rgba(255, 129, 130, 0.4);
   text-decoration: none;
 }
+
+.doc-diff-chunk-selected {
+  background-color: lightyellow;
```
agjohnson, line 15: You should specify an a value with an alpha, ie `rgba`. If you use `lightyellow` it will likely be illegible on a dark theme -- white text on lightyellow.

src/filetreediff.js
```diff
@@ -62,6 +105,21 @@ export class FileTreeDiffElement extends LitElement {
     return `${currentOrigin}${currentPath}`;
   }
 
+  renderArrows() {
+    if (!this.docDiffEnabled) {
+      return nothing;
+    }
+    console.log("chunkIndex", this.chunkIndex);
+    console.log("chunks.length", this.chunks.length);
+    console.log("chunks", this.chunks);
```
agjohnson, line 114: Reminder, we have `console.debug` for debug information. The console gets spammy with debug level information in it.

humitos: Good reminder, yeah. I should delete all of these, tho. I was using them while developing.

humitos:
I updated this PR with a more polished version. It looks great to me! 💯

humitos:
[Peek 2025-02-04 11-08.webm](https://github.com/user-attachments/assets/651870ab-a6bd-429d-899c-780472151bc6)

src/filetreediff.css
```diff
@@ -57,3 +57,13 @@
...
+
+:host > div > div > span {
+  display: flex;
+  width: 0.5rem;
+  cursor: pointer;
+}
+
+:host > div > div > span:nth-child(2) {
+  min-width: 2.5rem;
+}
```
ericholscher, line 69: This feels like it might be brittle in the future, and I think is an argument for using class names?

src/filetreediff.js
```diff
@@ -24,6 +30,41 @@ export class FileTreeDiffElement extends LitElement {
...
+      ":has(.doc-diff-removed)",
+    ];
+
+    this.chunkSelector = [];
+    for (let selector of this.chunkTagSelector) {
+      for (let pseudo of this.chunkPseudoSelector) {
+        this.chunkSelector.push(`${selector}${pseudo}`);
+      }
+    }
+    this.chunkSelector = this.chunkSelector.join(", ");
```
ericholscher, line 57: I don't really understand what this code is doing. Seems like we're making a big list of possible elements?

src/filetreediff.js
```diff
@@ -73,11 +125,61 @@ export class FileTreeDiffElement extends LitElement {
...
+    }
+
+    const chunk = this.chunks[this.chunkIndex - 1];
+    this.scrollToChunk(chunk);
+  }
+
+  nextChunk() {
+    if (!this.chunks.length) {
+      return;
+    }
```
ericholscher, line 151: We should probably disable the button in this case?

ericholscher (approved):
This looks good to me and fine to ship and iterate more 💯

src/filetreediff.js
```diff
@@ -73,11 +125,61 @@ export class FileTreeDiffElement extends LitElement {
...
+    if (this.chunkIndex != this.chunks.length) {
+      this.chunkIndex += 1;
+    }
+
+    const chunk = this.chunks[this.chunkIndex - 1];
```
ericholscher, line 156: Seems this code is the same for previous & next? Shouldn't this be adding a chunk?

```suggestion
    const chunk = this.chunks[this.chunkIndex + 1];
```

src/filetreediff.css
humitos: Good point. I added a `.chunks` class for the `1 of 8` text.

src/filetreediff.js
ericholscher: It would just grey out the button and not make it clickable. Realistically we should probably just hide the arrows if there's 0 chunks.

humitos: > Realistically we should probably just hide the arrows if there's 0 chunks.

I thought about that as well, but it feels it doesn't work when you click on "Show diff" and nothing happens. Maybe some copy could work better like: "No differences found" or something similar, but shorter 😄

PR #520, opened 2025-02-04
Docdiff: use Embed API to grab the base URL content

humitos:
Fetching a documentation URL is blocked because of CORS in Read the Docs for Business, so we use Embed API to fetch the content we need to perform the docdiff.

Closes https://github.com/readthedocs/readthedocs-corporate/issues/1942

### DIFF
--- a/src/docdiff.js
+++ b/src/docdiff.js
@@ -20,6 +20,7 @@ import {
 import { nothing, LitElement } from "lit";
 import { default as objectPath } from "object-path";
 import { hasQueryParam, docTool } from "./utils";
+import { EMBED_API_ENDPOINT } from "./constants";
 
 export const DOCDIFF_URL_PARAM = "readthedocs-diff";
 
@@ -84,7 +85,7 @@ export class DocDiffElement extends LitElement {
     this.injectStyles = true;
 
     this.originalBody = null;
-    this.cachedRemoteContent = null;
+    this.cachedRemoteResponse = null;
   }
 
   loadConfig(config) {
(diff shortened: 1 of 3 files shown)


src/docdiff.js
```diff
@@ -113,47 +114,43 @@ export class DocDiffElement extends LitElement {
...
-  //     this.disableDocDiff();
-  //   }
-  // }
+  getEmbedURL(url) {
+    const params = {
+      url: url,
+    };
+
+    if (this.rootSelector !== null) {
+      params["maincontent"] = this.rootSelector;
```
ericholscher, line 125: I'm a little worried we might miss content if we pass a root selector here that isn't in the page -- I don't think we've validating that it exists, and so might not return a proper value.

I guess if it isn't in the page the downstream parsing logic won't work either, so seems fine to query it here?

ericholscher (approved):

src/docdiff.js
humitos: Exactly. We use the same `this.rootSelector` in the docdiff code. If it doesn't exist, docdiff won't work either way.

ericholscher: Should we not issue the request if it doesn't exist, then?

humitos: Yeah, probably. We could check if the `rootSelector` exists in the current page before making the request. I'll open an issue for this.

humitos: Issue at https://github.com/readthedocs/addons/issues/521

PR #523, opened 2025-02-05
Filetreediff: allow users to ignore some files

humitos:
Filter the added/modified list of files to filter our "ignored files" coming from the API response.

[Peek 2025-02-05 14-26.webm](https://github.com/user-attachments/assets/f6e45326-fead-4f40-8b76-36a82b1d6cb0)

Requires readthedocs/readthedocs.org#11977


humitos:
This PR is not required anymore since the filtering is performed in the server now.

PR #526, opened 2025-02-10
File Tree Diff: style to match flyout

humitos:
This is just a test to see how it looks following the flyout look and feel.

Screenshot_2025-02-10_11-21-31

### DIFF
--- a/src/filetreediff.css
+++ b/src/filetreediff.css
@@ -61,7 +61,6 @@
 
 :host > div > div > label > input {
   margin: 0;
-  background-color: #ccc;
 }
 
 :host > div > div > span:not(.chunks) {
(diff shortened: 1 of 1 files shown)


ericholscher:
Looks reasonable. I wonder if there's a nicer way to have these colors defined more globally, and users can override them with a single variable?

humitos:
> I wonder if there's a nicer way to have these colors defined more globally, and users can override them with a single variable?

I suppose we will expose CSS variables like `--readthedocs-font-color` (without mentioning the addon) and use that as default for `--readthedocs-flyout-font-color`.

Besides, @agjohnson talked about using `@context` some time ago that we can use to define "themes" or these colors in a more global way.

ericholscher (approved):
:+1: if want to ship it.

humitos:
> I don't love the contrast of the grey text on the dark BG tho.

Yeah, that gray color is from flyout section. We can use a "whiter color" we also use in the flyout for the items of those sections. I looks like this:


Screenshot_2025-02-12_12-02-03

Maybe better?

PR #527, opened 2025-02-10
File Tree Diff: simplify the algorithm to find chunks

humitos:
- We get all the elements with `.doc-diff-*` classes
- For each element, we find the immediate parent matching a "section node"
- De-duplicate those nodes (multiple changes in the same section)

### DIFF
--- a/src/filetreediff.js
+++ b/src/filetreediff.js
@@ -33,31 +33,6 @@ export class FileTreeDiffElement extends LitElement {
 
     this.chunkIndex = 1;
     this.chunks = [];
-    this.chunkTagSelector = [
-      // We may want to add more selectors here as we find them.
-      "section",
-      "h1",
-      "h2",
-      "h3",
-      "p",
-      "dl",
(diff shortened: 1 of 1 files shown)


src/filetreediff.js
```diff
@@ -240,10 +215,72 @@ export class FileTreeDiffElement extends LitElement {
...
+    //     deleted/added inside a paragraph), we will return its parent `p`.
+    //   - If the class is added to a `section`, we will return the same
+    //     `section` element.
+    //   - If the class is added to a `li`, we will return the same `ul`/`ol`
+    //     element.
+    const sectionNodes = [
+      "section",
+      "h1",
+      "h2",
+      "h3",
```
ericholscher, line 239:
```suggestion
      "h3",
      "h4",
      "h5",
      "h6",
```

Seems reasonable for completeness?

ericholscher (approved):
This seems like a better approach for sure.

PR #532, opened 2025-02-12
Resolver filename fixes

humitos:
Make flyout work when `readthedocs.resolver.filename` is `null`. In that case, we get this value from the META tag added by Cloudflare.

This was the original intent here, but I missed allowing `null` in the data validation. Besides, even if the data validation allows `null`, there was another bug since I was re-defining the variable (`const ...`) instead of overriding its value.

We didn't catch this locally because we have a different rule that makes the request to always contain `url=` parameter and that case is not affected by this.

Besides, I didn't find this when doing QA after deploy because I tested with examples that sent `url=` parameter.

I'll see if I can write a test for this in another PR.

### DIFF
--- a/src/data-validation.js
+++ b/src/data-validation.js
@@ -250,7 +250,7 @@ const addons_flyout = {
           type: "object",
           required: ["filename"],
           properties: {
-            filename: { type: "string" },
+            filename: { type: ["string", "null"] },
           },
         },
       },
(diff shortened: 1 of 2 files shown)


humitos:
> I'll see if I can write a test for this in another PR.

I ended up pushing the tests to this PR 👍🏼

agjohnson (approved):
Test failure looks maybe unrelated, this looks good otherwise

humitos:
yeah, the tests are fixed in another PR.

PR #538, opened 2025-02-17
Links preview: cache requests to API

humitos:
Keep a cache of all the API request calls done and re-use them when the same link is previewed again. It follows the same logic we use on docdiff.

### DIFF
--- a/src/linkpreviews.js
+++ b/src/linkpreviews.js
@@ -26,6 +26,11 @@ const HIDE_TOOLTIP_DELAY = IS_TESTING ? 0 : 300;
 
 const TOOLTIP_DATA_HREF = "data-linkpreview-href";
 
+// Object to save the responses from the EmbedAPI.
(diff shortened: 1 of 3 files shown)


humitos:
```
 ❌ Search tests > show the modal and perform a query
      AssertionError: Snapshot Search tests show the modal and perform a query does not match the saved snapshot on disk
      + expected - actual
      
       <div role="search">
         <div class="background">
         </div>
         <div class="content">
      -    <form>
      +    <form class="focus">
             <label>
             </label>
             <input
               autocomplete="off"
```

It seems these tests behaves a little different in CircleCI than in local testing. I'll see if I find a way to ignore this. I tried https://open-wc.org/docs/testing/semantic-dom-diff/ but it didn't work for some reason.

humitos:
I seems I need to re-run the snapshots so that class is removed from the resulting snapshot 👍🏼

ericholscher (approved):

PR #539, opened 2025-02-18
FTD: Reduce width of full size display

ericholscher:


src/filetreediff.css
```diff
@@ -76,7 +76,8 @@
   min-width: 2.5rem;
 }
 
-@media (max-width: 1023px) {
+/* Expand to full width on smaller screens */
+@media (max-width: 700px) {
```
agjohnson, line 80: 768px is commonly used to target most small tablets and phones. You can review common breakpoints here:

https://www.browserstack.com/guide/responsive-design-breakpoints

ericholscher: Yea, I looked at our existing code, and didn't see a common number, so just picked one.

ericholscher:
```suggestion
@media (max-width: 768px) {
```

Issue #420, opened 2024-11-04
Search: Input text is white on white on dark mode


mhilbrunner:
Search: Input text is white on white on dark mode OS/browsers. It also seems the `--readthedocs-search-color` CSS var isn't actually set for us on https://docs.godotengine.org/en/latest, so this may be an issue on our end?

For now, I've disabled the search addon due to https://github.com/readthedocs/addons/issues/419.

ericholscher:
@humitos @zanderle This seems like something to prioritize if we can.

humitos:
We don't currently support dark mode by default yet (e.g. it's not possible to set a CSS class to make search modal dark). However, we do have a bunch of `--readthedocs-search-*` variables already in the code, and we are planning to document this type of customization more. See https://github.com/readthedocs/readthedocs.org/pull/11888

@mhilbrunner @rjwills28  If you are not able to make dark mode working properly with the current set of `--readthedocs-search-*` variables; what other values/variables you would need to change to make it work as you expected? That will help us to expose these variables to users so they can customize the look and feel.

rjwills28:
Thanks for the reply @humitos  - I was able to come up with a workaround for this issue but will also have a look at the `--readthedocs-search-*` variables to see if that is cleaner.

ericholscher:
We're going to start a basic implementation of a dark mode here to start.

agjohnson:
What I mentioned in the meeting is a mix of two patterns. We'd still use variable underneath, but we don't require theme maintainers or users to set all of these variables to achieve a dark mode.

```css
@layer defaults {
  :root[data-theme="light"],
  :root:not([data-theme="dark"]) {
    color-scheme: light;
    --readthedocs-flyout-background-color: ...;
    --readthedocs-notification-background-color: ...;
    ...
  }

  :root[data-theme="dark"] {
    color-scheme: dark;
    --readthedocs-flyout-background-color: ...;
    --readthedocs-notification-background-color: ...;
    ...
  }
}
```

Just as an example, this specifies the dark mode as an attribute on `html`. The individual tool would be responsible for setting this attribute at least to start. There are other selectors we could use here too, but the aim is to give reader user controllable dark mode, defaulting to light mode.

Using `prefers-color-scheme` alone is not enough as the underlying tool might not support dark mode. But we can use the `color-scheme` meta to trigger usage of `prefers-color-scheme`:

```html
<meta name="color-scheme" content="dark light" />
```

[...]

PR #540, opened 2025-02-18
Force black text for search input

ericholscher:
Fixes https://github.com/readthedocs/addons/issues/420

Screenshot 2025-02-18 at 2 14 53 PM
Screenshot 2025-02-18 at 2 15 00 PM

### DIFF
--- a/src/search.css
+++ b/src/search.css
@@ -114,7 +114,7 @@
   width: 100%;
   padding: 6px;
   line-height: 1;
-  color: black; /* Fix https://github.com/readthedocs/addons/issues/420 */
+  color: var(--readthedocs-search-color, rgb(64, 64, 64));
 }
 
 :host > div .results {
--- a/src/search.css
+++ b/src/search.css
@@ -34,9 +34,8 @@
   white-space: pre-wrap;
   word-wrap: break-word;
   font-size: 0.875em;
-  font-family:
-    SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New",
-    monospace;
+  font-family: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono",
+    "Courier New", monospace;
   border-radius: 0;
 }
 
@@ -114,6 +113,7 @@
   width: 100%;
   padding: 6px;
   line-height: 1;
+  color: black; /* Fix https://github.com/readthedocs/addons/issues/420 */
 }
 
 :host > div .results {
--- a/src/search.css
+++ b/src/search.css
@@ -34,8 +34,9 @@
   white-space: pre-wrap;
   word-wrap: break-word;
   font-size: 0.875em;
-  font-family: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono",
-    "Courier New", monospace;
+  font-family:
+    SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New",
+    monospace;
   border-radius: 0;
 }
 


ericholscher:
This lints fine for me locally 🤷

src/search.css
```diff
@@ -114,6 +114,7 @@
   width: 100%;
   padding: 6px;
   line-height: 1;
+  color: black; /* Fix https://github.com/readthedocs/addons/issues/420 */
```
agjohnson, line 117: This should use a variable instead of hard coding the value. I looked into this and the issue is that the computed value is from the element style, which has higher specificity than the inherited value from the parent div. So reiterating the inherited value here is probably best either way.

image


```suggestion
  color: var(--readthedocs-search-color, rgb(64, 64, 64));
```

humitos:
You should probably do the same work for the "X" shown at the right of the input field. It's currently white as well.

src/search.css
```diff
@@ -114,6 +114,18 @@
...
+  color: var(--readthedocs-search-color, rgb(64, 64, 64));
+}
+
+/* Override the search input X button with custom black svg */
+/* Via https://stackoverflow.com/a/68759280 */
+[type="search"]::-webkit-search-cancel-button {
+  -webkit-appearance: none;
+  appearance: none;
+  height: 10px;
+  width: 10px;
```
agjohnson, line 126: Use `em` instead of `px` here, `px` won't scale with changes to the font size.

src/search.css
```diff
@@ -114,6 +114,18 @@
...
+  background-size: 10px 10px;
```
agjohnson, line 128: Same on `em` here

src/search.css
```diff
@@ -114,6 +114,18 @@
...
...NWRyAxLjEvL0VOIiAiaHR0cDovL3d3dy53My5vcmcvR3JhcGhpY3MvU1ZHLzEuMS9EVEQvc3ZnMTEuZHRkIj4NCjxzdmcgdmVyc2lvbj0iMS4xIiBpZD0iQ2FwYV8xIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHhtbG5zOnhsaW5rPSJodHRwOi8vd3d3LnczLm9yZy8xOTk5L3hsaW5rIiB4PSIwcHgiIHk9IjBweCINCgkgd2lkdGg9IjEyMy4wNXB4IiBoZWlnaHQ9IjEyMy4wNXB4IiB2aWV3Qm94PSIwIDAgMTIzLjA1IDEyMy4wNSIgc3R5bGU9ImVuYWJsZS1iYWNrZ3JvdW5kOm5ldyAwIDAgMTIzLjA1IDEyMy4wNTsiDQoJIHhtbDpzcGFjZT0icHJlc2VydmUiPg0KPGc+DQoJPHBhdGggZD0iTTEyMS4zMjUsMTAuOTI1bC04LjUtOC4zOTljLTIuMy0yLjMtNi4xLTIuMy04LjUsMGwtNDIuNCw0Mi4zOTlMMTguNzI2LDEuNzI2Yy0yLjMwMS0yLjMwMS02LjEwMS0yLjMwMS04LjUsMGwtOC41LDguNQ0KCQljLTIuMzAxLDIuMy0yLjMwMSw2LjEsMCw4LjVsNDMuMSw0My4xbC00Mi4zLDQyLjVjLTIuMywyLjMtMi4zLDYuMSwwLDguNWw4LjUsOC41YzIuMywyLjMsNi4xLDIuMyw4LjUsMGw0Mi4zOTktNDIuNGw0Mi40LDQyLjQNCgkJYzIuMywyLjMsNi4xLDIuMyw4LjUsMGw4LjUtOC41YzIuMy0yLjMsMi4zLTYuMSwwLTguNWwtNDIuNS00Mi40bDQyLjQtNDIuMzk5QzEyMy42MjUsMTcuMTI1LDEyMy42MjUsMTMuMzI1LDEyMS4zMjUsMTAuOTI1eiIvPg0KPC9nPg0KPGc+DQo8L2c+DQo8Zz4NCjwvZz4NCjxnPg0KPC9nPg0KPGc+DQo8L2c+DQo8Zz4NCjwvZz4NCjxnPg0KPC9nPg0KPGc+DQo8L2c+DQo8Zz4NCjwvZz4NCjxnPg0KPC9nPg0KPGc+DQo8L2c+DQo8Zz4NCjwvZz4NCjxnPg0KPC9nPg0KPGc+DQo8L2c+DQo8Zz4NCjwvZz4NCjxnPg0KPC9nPg0KPC9zdmc+DQo=);
```
agjohnson, line 127: Hrm, it's unfortunate that this behavior is webkit specific. I think it might be better to match the UX/UI on all browsers and just hide this in webkit for now. Hardcoding images in our CSS won't be very maintainable, nor customizable by users. If we want this feature in the search input, it seems like we should use the pattern we're using and add the icon via Font Awesome and JS. This also allows users to easily customize via font style properties.

humitos: > it seems like we should use the pattern we're using and add the icon via Font Awesome and JS

👍🏼

humitos: @ericholscher I don't want to block this PR because of the X icon, so I'd suggest to revert the change you've done here, open an issue for the X bug and move forward with the fix on the text input.

ericholscher:
Reverted.

humitos:
BTW, the X works fine in Chrome to me. It's not shown on Firefox.

Screenshot_2025-02-25_17-08-39

ericholscher:
It shows white for me in Chrome, but at least you can read the text 🤷

Issue #541, opened 2025-02-18
Flyout gets reset in odd state after search is opened


ericholscher:
When you open search from the flyout, it minimizes it, but in a way that was different than it should be:

Before:

![Image](https://github.com/user-attachments/assets/1cda380f-d9ff-441e-bff0-991a640ca1e7)

After:

![Image](https://github.com/user-attachments/assets/abdc72e4-da02-4bb0-9d34-6b21ab7d8d13)

humitos:
Closing as this will be re-designed completely in https://github.com/readthedocs/addons/issues/574

PR #542, opened 2025-02-18
Use single method to open/close flyout

ericholscher:
Fixes https://github.com/readthedocs/addons/issues/541

### DIFF
--- a/src/flyout.js
+++ b/src/flyout.js
@@ -366,11 +366,11 @@ export class FlyoutElement extends LitElement {
   }
 
   _showFlyout = (e) => {
-    this.opened = true;
+    this._open();
   };
 
   _hideFlyout = (e) => {
-    this.opened = false;
+    this._close();
   };
 
   connectedCallback() {


src/flyout.js
```diff
@@ -366,11 +366,11 @@ export class FlyoutElement extends LitElement {
   }
 
   _showFlyout = (e) => {
-    this.opened = true;
+    this._open();
   };
 
   _hideFlyout = (e) => {
-    this.opened = false;
+    this._close();
```
agjohnson, line 373: Lit provides a native way to update dependent attributes/variables on reactive property changes. We should use that pattern instead of relying on manually calling methods like this. With `willUpdate()`, the current code would be fine

https://lit.dev/docs/components/lifecycle/#willupdate

humitos (commented):
Take a look at the tests I wrote and try to write a test case for this case. These cases where we need to interact with the elements to do QA are pretty hard to notice after deploy otherwise.

ericholscher:
Cool... I'm a little lost in general with all of this, so might take a while to get back to this.

humitos:
@ericholscher let me know if I can help here or if you want me to take over it.

ericholscher:
Feel free to run with it. It was just a small bug I wanted to fix, but the PR review seems to require a ton of research that I don't have energy for currently.

humitos:
This is still a bug and Anthony opened a bye issue for it. It's not that we don't want to ship it, there is feedback about how to use the framework to solve this issue that wasn't addressed: https://github.com/readthedocs/addons/pull/542#discussion_r1960803659

ericholscher:
Yea, it's definitely the same issue as #585.

PR #543, opened 2025-02-19
Match notification style with flyout

ericholscher:
Refs https://github.com/readthedocs/ext-theme/pull/564#issuecomment-2657003600

Screenshot 2025-02-18 at 4 23 27 PM

Inverted:

Screenshot 2025-02-18 at 4 38 45 PM

### DIFF
--- a/src/notification.css
+++ b/src/notification.css
@@ -3,14 +3,8 @@
     --readthedocs-notification-font-size,
     var(--readthedocs-font-size, 1rem)
   );
-  --addons-notification-color: var(
-    --readthedocs-notification-color,
-    rgb(252, 252, 252)
-  );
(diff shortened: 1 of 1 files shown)


ericholscher:
I wasn't sure if we were currently using `inverted` at all?

agjohnson:
I was saying in our meeting that we shouldn't yet match the light theme elements against the flyout. Same with the diff UI. These are dark mode themes and they stand out a great deal on light mode themes

humitos:
I think we should probably close this branch and put the effort in the new version of the toolbar (discussed at the offsite) that will include the notifications as well.

ericholscher:
Closing this, since it sounds like we don't want to move forward on it.

PR #544, opened 2025-02-19
Fix bug with ready event

agjohnson:
The `return` here is incorrect and this is causing the function to be
exiting too early.


humitos (approved):
Works fine. It would be good if you can write a test case for this.

agjohnson:
I was surprised there were no tests to catch this bug. I won't have much time to dive deeper here though and don't want to hold this up from merging. I'll add an issue though

PR #545, opened 2025-02-19
Try CSS variable again

agjohnson:
Potentially replaces #492


public/_/readthedocs-addons.json
agjohnson, line 1: This file is for illustration only.

public/global.css
agjohnson, line 1: Also for illustration only

public/index.html
agjohnson, line 1: And also not needed

agjohnson:
Marking this for review so it doesn't get lost again. @humitos can you look at this and see if it makes sense to you?

There are some extra changes here that can be removed before merging, it's just for example.

Right now, with all of the changes and the defaults applied with the extra files, here is what the flyout and notification look like:

image

src/defaults.js
```diff
@@ -1,42 +1,55 @@
...
+export const defaultStyleSheet = new CSSStyleSheet();
+
 /*
  * Specific styles based on documentation tools and themes
  *
  * Usage of `@layer` at-rule pushes this rules down a step in
  * precedence/priority. This allows a user `:root` rule to override these
  * values.
  **/
+defaultStyleSheet.replaceSync(`
```
humitos, line 15: I don't like too much that we are writing all these CSS in a JS file. Can't we just have a `defaults.css` file and import it here instead?

src/defaults.js
```diff
@@ -1,42 +1,55 @@
...
 
   :root[data-readthedocs-tool="sphinx"][data-readthedocs-tool-theme="furo"] {
     --readthedocs-font-size: 0.725rem;
-    --readthedocs-flyout-header-font-size: 0.845rem;
     --readthedocs-flyout-font-size: 0.725rem;
   }
 
   :root[data-readthedocs-tool="sphinx"][data-readthedocs-tool-theme="immaterial"] {
     --readthedocs-font-size: 0.58rem;
-    --readthedocs-flyout-header-font-size: 0.7rem;
```
humitos, line 39: Why did you remove all the `flyout-header-font-size`?

public/index.html
humitos: We can probably create another HTML file to keep in the repository and show this as example 👍🏼

src/defaults.js
humitos: Actually, I don't really understand why you migrated from `doctools.css` to `default.js`. You had to change the code in `application.js` as well -- why not just keep the `doctools.css` (or `defaults.css` if you want) but keeping the CSS extension and importing the CSS from `application.js` as we were doing?

src/flyout.css
```diff
@@ -1,23 +1,35 @@
...
-    --readthedocs-flyout-font-size,
-    var(--readthedocs-font-size, 0.8rem)
-  );
-  --addons-flyout-line-height: var(--readthedocs-flyout-line-height, 1.25em);
+@layer defaults {
+  :root {
+    --readthedocs-flyout-max-width: 25em;
+
+    --readthedocs-flyout-font-family: var(--readthedocs-font-family);
+    --readthedocs-flyout-font-size: var(--readthedocs-font-size);
```
humitos, line 8: It seems we need to define these in `defaults.js`, right? I mean, `--readthedocs-font-*`. I didn't find where they are defined.

src/flyout.css
```diff
@@ -80,7 +85,9 @@ header > img.logo {
 }
 
 header > span {
-  color: var(--readthedocs-flyout-current-version-color, #27ae60);
+  color: var(--readthedocs-flyout-current-version-color);
+  font-size: var(--readthedocs-flyout-header-font-size);
+  margin-left: 10px;
```
humitos, line 90: This property is new. Is this correct?

src/flyout.css
```diff
@@ -126,25 +133,19 @@ dl {
...
+  color: var(--readthedocs-flyout-section-heading-color);
 }
 
 dl > dd {
   display: inline-block;
   margin: 0;
-  font-size: var(
-    --readthedocs-flyout-dd-font-size,
-    calc(var(--addons-flyout-font-size) * 1.125)
-  );
```
humitos, line 142: We are missing the calculated value here.

humitos (commented):
I think this is a good pattern and it's going in a good direction.

Here you have my observations:

* I'd like to write the CSS code in a CSS file
* We should have a new HTML file showing an example that overrides the defaults
* There are some properties that were calculated based in others and now they are fixed values
* Some defaults are not defined (eg. font family)

src/defaults.js
agjohnson: That should be computed from font-size using a relative unit already. It did not need to be a static unit.

src/flyout.css
agjohnson: This value comes from a relative value using `em` instead of using `calc()` to accomplish this.

src/defaults.js
agjohnson: It can't be an imported CSS file (Lit stylesheet) as these lack all of the functionality of the native CSSStyleSheet instances, namely `insertRule()` used below.

So I would not make this a CSS import, what is here is easiest. There are only a few lines of CSS here, so this doesn't feel like a barrier. If we eventually find we would gain some value from splitting this out to another CSS file, we can create a merge strategy that includes a separate file first.

src/flyout.css
agjohnson: Yup correct. This was probably lost in merging.

flying-sheep:
See here for some currently hardcoded colors: https://github.com/readthedocs/addons/issues/570#issuecomment-2820945727

Here’s an adapter for the pydata theme you can try:

```css
:root {
    --readthedocs-search-backdrop-color: var(--pst-color-shadow);
    --readthedocs-search-color: var(--pst-color-text-base);
    --readthedocs-search-content-background-color: var(--pst-color-background);
    --readthedocs-search-content-border-color: var(--pst-color-border);
    --readthedocs-search-filters-border-color: var(--pst-color-border);
    --readthedocs-search-font-family: var(--pst-font-family-base);
    --readthedocs-search-font-size: var(--pst-font-size-base);
    --readthedocs-search-footer-background-color: var(--pst-color-on-background);
    --readthedocs-search-footer-code-background-color: var(--pst-color-surface);
    --readthedocs-search-footer-code-border-color: var(--pst-color-border);
    --readthedocs-search-input-background-color: var(--pst-color-surface);
    --readthedocs-search-result-section-border-color: var(--pst-color-border);
    --readthedocs-search-result-section-color: var(--pst-color-link);
    --readthedocs-search-result-section-highlight-color: var(--pst-color-accent);
    --readthedocs-search-result-section-subheading-color: var(--pst-color-text-muted);
}
```

agjohnson:
> See here for some currently hardcoded colors: https://github.com/readthedocs/addons/issues/570#issuecomment-2820945727

I had skipped the search addon for this first pass, but it probably makes sense to also use these changes here. I'll replicate these variables in this PR :+1:

humitos (approved):
I re-review this and I think the pattern is correct 👍🏼 

The last 3 bullets from my previous review are still valid: https://github.com/readthedocs/addons/pull/545#pullrequestreview-2644314426

agjohnson:
> We should have a new HTML file showing an example that overrides the defaults

You can add this on the existing example test page as a rule on `html`. I don't think we need to duplicate the test page just to show this.

> There are some properties that were calculated based in others and now they are fixed values

What are you referring to here? There are a few places that a complex `calc(var(...) * 1.125)` was replaced by a much simpler `1.125em` but I'm finding

> Some defaults are not defined (eg. font family)

Also what are you referring to here?

agjohnson:
I am still poking this as the search addon is quite chaotic on font-sizes. It overuses `font-size` and uses all sorts of different units are used for sizing and positioning: rem, em, px, `small`, and percentage. So the font size is customizable but it renders inconsistently:

image

image

humitos:
> What are you referring to here? There are a few places that a complex `calc(var(...) * 1.125)` was replaced by a much simpler `1.125em`, is this what you're referring to?

Yes, I mentioned this at https://github.com/readthedocs/addons/pull/545#discussion_r1971519767

It seems you already replied to that and I've missed.

agjohnson:
I wrote up a quick example of what CSS style testing would look like in 4b683a3. I didn't do this across our addons or customization points yet, but this would be easy to expand on.

I'm going to push off fixing issues with font sizes in search to a later issue. The font size issue is a current bug with search already:

image

So I think we can continue with this from here. Follow up work to this issue can be:

- [ ] Test this in production more
- [ ] Tests: improve coverage on CSS customization unit tests
- [ ] Search: fix inconsistent and static font size usage
- [ ] #577 to document customization and defaults better

src/defaults.js
agjohnson: I just ended up moving this to a separate CSS file, skipping Lit for these stylesheet entirely.

src/search.css
```diff
@@ -1,16 +1,57 @@
...
+    --readthedocs-search-content-background-color: #fcfcfc;
+    --readthedocs-search-content-border-color: #e0e0e0;
+    --readthedocs-search-filters-border-color: #e0e0e0;
+    --readthedocs-search-footer-background-color: rgb(234, 234, 234);
+    --readthedocs-search-footer-color: #404040;
+    --readthedocs-search-footer-code-background-color: #fff;
+    --readthedocs-search-footer-code-border-color: #e1e4e5;
+    --readthedocs-search-input-background-color: #eaeaea;
+
+    /* `--readthedocs-search-result-section-highlight-color` is deprecated but
```
agjohnson, line 23: The `result-section` prefix confused me, as these rules target not the "section" -- ie the surrounding element around results -- but are instead for the individual results. So I think these should all be `search-result-background-color` not `search-result-section-background-color`, etc.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v3.1.0"
+  - repo: https://github.com/rbubley/mirrors-prettier
+    rev: "v3.5.3"
```
agjohnson, line 3: I'm still noticing a lot of conflict between just running `npm run format` and running prettier through pre-commit. I feel like we should run `npm run format` instead.

agjohnson:
@humitos I'd take one last look at the change here. Notable changes at this point:

- Slight CSS unit test coverage. Anything else to cover immediately?
- Not a big deal, but I noticed that tests do need to load the default style sheets manually now. Without loading the defaults, unit tests load partially unstyled elements.
- Some CSS variables have changed slightly

humitos:
> We can unit test some of these styles with web-test-runner though, and should probably eventually use a visual diffing tool like Happo to just automate this.

Yeah, no, I'm not talking about something complex like this. I was thinking of just having a page _using_ the CSS variables with noticeable changes (very big flyout/search/etc) just to quickly see they have effect.

humitos:
> I am still poking this as the search addon is quite chaotic on font-sizes. It overuses `font-size` and uses all sorts of different units are used for sizing and positioning: rem, em, px, and `small`.

This was quickly copied from the old Sphinx extension without thinking too much about it or doing any modification. Feel free to re-structure it with a more modern pattern.

.pre-commit-config.yaml
humitos: Hrm, weird... Maybe it's because of different versions? 🤔 . I don't recall hitting this issue.

src/notification.css
```diff
@@ -74,63 +79,57 @@
...
+  line-height: 1em;
   font-weight: bold;
 }
 :host > div > div.content {
-  line-height: var(--addons-notification-font-size);
-  font-size: calc(var(--addons-notification-font-size) * 0.85);
-  padding: 0 calc(var(--addons-notification-font-size) * 1.5);
+  line-height: 1em;
+  font-size: 0.85em;
+  padding: 0 1.5em;
```
humitos, line 118: I don't understand some of these changes. Previously they were based on a variable and now they are fixed values. Is this because they are still relative units (em)? However, they were relative to `font-size` previously and now... relative to...

> The em unit is relative to its nearest parent element's font size

(from Google)

so, I understand it's the same as before 😄

humitos: Changing `font-size` should change all these values automatically... wow, CSS 🤯

src/search.css
humitos: I don't remember exactly how this works, but I remember that a section can have multiple results -- if that helps 😄

tests/flyout.test.html
```diff
@@ -355,6 +356,67 @@
...
         });
+
+        describe("Flyout element styling", () => {
+          it("supports font size customization with fallback", async () => {
+            document.adoptedStyleSheets = [defaultStyleSheet];
+            const addon = new flyout.FlyoutAddon(config);
+            const element = document.querySelector("readthedocs-flyout");
+            await elementUpdated(element);
+
+            // Addon default of 80% of global --readthedocs-font-size, default 16px
```
humitos, line 367: Where this "default 16px" comes from? We are defining these values in `em`, how they are converted into `px` at start up?

tests/flyout.test.html
```diff
@@ -355,6 +356,67 @@
             );
           });
         });
+
+        describe("Flyout element styling", () => {
+          it("supports font size customization with fallback", async () => {
+            document.adoptedStyleSheets = [defaultStyleSheet];
```
humitos, line 362: It would be good to put a comment here about what you found. I'm sure we will be confused about _why_ we need to load this here manually.

humitos (approved):

src/defaults.css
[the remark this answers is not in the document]
agjohnson: I updated this to use a different repo for now, I was getting different results locally and in pre-commit. The pre-commit repo stated it's archived but it really vague on why.

https://github.com/pre-commit/mirrors-prettier

> archived
> prettier made some changes that breaks plugins entirely

:shrug:

src/notification.css
agjohnson: Yes, same as before unless another `font-size` definition is added to an element somewhere in the hierarchy.

We did waffle on this a bit but I lean towards using `em` inside web components for simplicity. `rem` is best outside the shadow DOM but inside it always refers to the host `rem` and so is not useful. The main reason `rem` is useful is because pages tend to have a lot of nesting and so a lot of potential `font-size` is accidentally redeclared. Web components are granular UI and so have less potential for all of this.

Using `calc` does work but gets really verbose for rather basic geometry needs. I might still use `calc` if there is a lot of nesting and font-size changes.

tests/flyout.test.html
agjohnson: This is a requirement. We're using the elements directly and so skipping all of the set up logic, including setting the default styles.

.pre-commit-config.yaml
agjohnson: Pre-commit was only installing 3.1 and 3.5 was installed locally. I don't want to pin prettier locally based on what pre-commit supports as the prettier plugin seems to be archived anyways.

src/search.css
agjohnson: There is a `div.results` that has all of the individual page `.hit` elements and each `.hit` can have multiple nested results. But the styles in these variables aren't affecting `div.results`, what I assumed `section` referred to. It's ambiguous enough it's worth removing either way.

humitos:
Should we merge and deploy this?

tests/flyout.test.html
agjohnson: From the browser, 16px is the default font size for almost all browsers.

agjohnson:
I'm fine merging this but I wouldn't deploy this tomorrow if we aren't all available. That's fine though, we can just release a day that we are around.

PR #548, opened 2025-02-19
Remove PR build notification and add info to Visual Diff

ericholscher:
This removes the PR build notification,
and includes some information in the Visual Diff UI with similar information.

I tested with a real version, but these would be PR #'s in normal usage:

Screenshot 2025-02-19 at 3 48 39 PM

Refs https://github.com/readthedocs/meta/issues/176

### DIFF
--- a/src/filetreediff.js
+++ b/src/filetreediff.js
@@ -212,7 +212,7 @@ export class FileTreeDiffElement extends LitElement {
                 Go to build log
               </option>
               <option value="${this.config.versions.current.urls.vcs}">
-                Go to #${this.config.versions.current.slug}
+                Go to pull request #${this.config.versions.current.slug}
               </option>
               <option
                 value="https://docs.readthedocs.com/platform/stable/visual-diff.html"
--- a/src/filetreediff.js
+++ b/src/filetreediff.js
@@ -11,6 +11,7 @@ import {
 } from "./events";
 import { getQueryParam } from "./utils";
 import { AddonBase } from "./utils";
+import { addUtmParameters } from "./utils";
 
 const SCROLL_OFFSET_Y = 0.1;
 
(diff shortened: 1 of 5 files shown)


ericholscher:
This is like a baby command palette :D

agjohnson:
It's odd to put unrelated links in this dropdown. The dropdown by default is going to have a selected value of `Files changed in #markdoc`, or have a filename after selecting. It's not going to be a discoverable place for users to find links to documentation/etc.

The more standard place to put this is under a kebab dropdown menu, or add the information circle icon with some sort of popup/etc.

I think what we talked about in the meeting was clickable element in the UI instead of repurposing the dropdown. I still like the default text addition to the dropdown, this adds enough context to drop the notification

src/filetreediff.js
```diff
@@ -197,10 +198,28 @@ export class FileTreeDiffElement extends LitElement {
           ${this.renderDocDiff()}
           <select @change=${this.handleFileChange}>
             <option value="" ?selected=${!hasCurrentFile} disabled>
-              Files changed
+              Files changed in #${this.config.versions.current.slug}
```
agjohnson, line 201: This needs to be conditional based on what the build is it seems. If it's a named branch, we should not use a `#` prefix, and if this is a pull request we want more descriptive `PR #1234` instead here. This makes it immediately recognizable that this is a pull request build, without expanding this menu.

src/filetreediff.js
```diff
@@ -197,10 +198,28 @@ export class FileTreeDiffElement extends LitElement {
...
             ${renderSection(diffData.added, "Added")}
             ${renderSection(diffData.modified, "Changed")}
+            <optgroup label="Links">
+              <option
+                value="${addUtmParameters(
+                  this.config.builds.current.urls.build,
+                  "notification",
+                )}"
+              >
+                Go to build log
```
agjohnson, line 212: The verb seems unneeded on these. These would also feel more natural under a dedicated menu, which would remove the need to explain that these are links that go to another page.

```suggestion
                Build log
```

humitos:
I agree with Anthony here. I would

- add a small ℹ️ icon that links to the docs in the right
- add a small `#1234` with the PR number before the icon
- don't add any unrelated links to the dropdown; only show the list of files
- do not remove the current notification

I understand that's what we agreed in our meeting.

humitos:
I opened a PR with these changes in https://github.com/readthedocs/addons/pull/551

PR #551, opened 2025-02-20
File Tree Diff: updates to UI (info and PR link)

### LINKED ISSUE #549: Visual diff: add a small icon to link to the docs
Image

 
* Related: https://github.com/readthedocs/meta/issues/176

humitos:
This is what I understood we agreed on in our meetting.

[Peek 2025-02-20 11-45.webm](https://github.com/user-attachments/assets/904062f9-aeb3-4662-abf7-6037a5e5e3bd)

* Closes #549
* Closes #548

## ToDo

- [x] add tests
- [x] require build URL in data validation


ericholscher (approved):
Yea, this is way nicer 💯 

It doesn't remove the PR build notification tho :D

humitos:
> Only if we have it off by default? I really think avoiding notifications unless totally necessary is the right UX. The visual diff UI is a much better way to communicate that info.

I can open another PR for this change but I need to understand what's the exact behavior you want here first.

Do you want to remove the notification " This page was created from a pull request build"...

Screenshot_2025-02-20_17-35-21


- completely/always?
- only when file tree diff is enabled?
- even if we merge #552?
- etc...

If you describe all the cases for the expected UX I can put a PR with that.

humitos:
Also, take into account that we have a setting in the WebUI: https://github.com/readthedocs/readthedocs.org/blob/29a912f4aad30b83b29539cf4ab46c5f7721f040/readthedocs/projects/models.py#L262

that controls whether or not show notification on external versions. I'm not sure we want to ignore that setting.

humitos:
Summarizing, I don't want to change the current behavior without considering all of that -- but I'm happy to talk and change the current behavior if it's not what we want.

ericholscher:
I want inline PR build notifications that are off by default, but can be enabled in the dashboard with that model field.

humitos:
> I want inline PR build notifications that are off by default, but can be enabled in the dashboard with that model field.

OK. That would need #552 and also to change our current default value in .org.

I will merge this PR and we can continue discussing that in another issue. I opened https://github.com/readthedocs/addons/issues/557 for this.

PR #559, opened 2025-02-25
Test: `readthedocs-addons-data-ready`

### LINKED ISSUE #547: Test: events around data ready are fired correcctly
The changes in

- #504

introduced a bug with data ready event handling. There was a errant return that caused the data ready events to not fire, which was fixed in #544. I was surprised that there were no tests that caught this bug, so it seems we need to test that both the internal and external data ready tests are fired correctly.

humitos:
Closes #547


humitos:
I'm going to merge these tests, but feel free to review them. I'm happy to open a new PR with the feedback received here.

PR #561, opened 2025-02-26
Visual diff: check for `rootSelector` before making request API

### LINKED ISSUE #521: Visual diff: check the `rootSelector` exists before making the EmbedAPI request
We perform visual diff over the `rootSelector` (default or defined by the user) at;

https://github.com/readthedocs/addons/blob/5b70166e015920f144839813ffbf89f6f67118fb/src/docdiff.js#L169-L172

If the `rootSelector` is not found in the current page, visual diff won't work. We can take advantage of this and do not perform the EmbedAPI request if we already know it's not gonna work 😄 

We will need to check this at https://github.com/readthedocs/addons/blob/5b70166e015920f144839813ffbf89f6f67118fb/src/docdiff.js#L142-L147

humitos:
Closes #521

### DIFF
--- a/src/docdiff.js
+++ b/src/docdiff.js
@@ -147,8 +147,13 @@ export class DocDiffElement extends LitElement {
   }
 
   compare() {
-    let promiseData;
+    // First check the root selector is in the current body
+    if (document.querySelector(this.rootSelector) === null) {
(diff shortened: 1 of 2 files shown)


agjohnson (approved):
Looks good to me :+1:

PR #562, opened 2025-03-07
Use footer text content ad for unknown themes

ericholscher:
This is much less intrusive than the stickybox,
and allows us to still have an ad injected,
but not cover the content for users when this happens.

### DIFF
--- a/src/ethicalads.js
+++ b/src/ethicalads.js
@@ -196,18 +196,18 @@ export class EthicalAdsAddon extends AddonBase {
         if (elementToAppend) {
           elementToAppend.append(placement);
         }
-      } else if (window.innerWidth > 1300) {
-        // https://ethical-ad-client.readthedocs.io/en/latest/#stickybox
-        placement.setAttribute("data-ea-type", "image");
-        placement.setAttribute("data-ea-style", "stickybox");
-        this.addEaPlacementToElement(placement);
-        // `document.body` here is not too much relevant, since we are going to
-        // use this selector only for a floating stickybox ad
-        const elementInsertBefore = document.body;
-        elementInsertBefore.insertBefore(
-          placement,
-          elementInsertBefore.lastChild,
-        );
+      } else {
+        // Default to a text ad appended to the root selector when no known placement found
+        placement.setAttribute("data-ea-type", "text");
+
+        const rootSelector = docTool.getRootSelector();
+        const rootElement = document.querySelector(rootSelector);
+
+        if (rootElement) {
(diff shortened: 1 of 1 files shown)


davidfischer (approved):
Looks good. My only worry is in the rare case where none of the selectors match and we just append the ad to `<body>` but that probably won't happen.

https://github.com/readthedocs/addons/blob/57d34df9117a3aaaf0f503299b72ab68d9f52964/src/utils.js#L400

src/ethicalads.js
```diff
@@ -196,18 +196,21 @@ export class EthicalAdsAddon extends AddonBase {
...
-        elementInsertBefore.insertBefore(
-          placement,
-          elementInsertBefore.lastChild,
-        );
+      } else {
+        // Default to a text ad appended to the root selector when no known placement found
+        placement.setAttribute("data-ea-type", "text");
+        // TODO: Check this placement on the dashboard,
+        // and see how this is performing.
+        placement.setAttribute("id", "readthedocs-text-footer");
```
humitos, line 204: Do we _need a different_ ID here? We are already using `readthedocs-ea` for all the ads. I would prefer to keep consistency here.

humitos (approved):
The code looks good. I haven't tested it locally, tho.

I would prefer to keep using the same ID as we are using for all the other ads, for consistency, tho.

src/ethicalads.js
ericholscher: Yea, the goal is to track stats to see how it's performing, so we need a unique ID.

humitos: Ah, OK. Makes sense.

Can we use `readthedocs-ea-text-footer` (note the extra `ea-`) to keep consistency with the other ID?

humitos: Also, if we want to get more detailed data about all the ads, we can use `readthedocs-ea-<doctool>-<ad-type>` for all of them so we can filter by doctool _and_ type_.

humitos:
This looks great! I tested it locally and I found only two doctools where it doesn't work great, but they seem to be special cases and we can adapt/fix them later in another PR. I think we can merge it and deploy it tomorrow.

### Pelican

Screenshot_2025-03-17_13-19-01

### Sphinx Awesome theme

Screenshot_2025-03-17_13-18-51

ericholscher:
Yea, I think we can adjust it a bit as we see how it works, and also of course just not add the ad if we fallback to `body` if it's an issue in practice.

PR #592, opened 2025-06-02
Link preview: do not add elements inside headers

### LINKED ISSUE #584: Link preview of a link inside an h2 incorrectly inherits h2 styling for preview content?
Hi! We just moved the Hypothesis documentation from `sphinx-hoverxref` to the Link Preview addon. After doing so, I noticed that some of the tooltips display poorly:

Image

https://hypothesis.readthedocs.io/en/latest/index.html

Where the link is inside an `h2`, and the `data-linkpreview-href` tooltip div is placed inside of that containing `h2` element, meaning the entire tooltip content inherits the h2's font size / font weight / etc styles. 

I checked other preview links inside headers, like [the `@given` link in this header](https://hypothesis.readthedocs.io/en/latest/reference/api.html#arguments-to-given), but those appended the tooltip to a root `readthedocs-linkpreviews` element and display fine, since they don't inherit the `h2` styles.

I'm not sure what is causing the behavior difference here, but I'm hoping you guys have some idea!

A confounding factor is that on the page where the tooltip div displays poorly, we have some custom css ([source](https://hypothesis.readthedocs.io/en/latest/_sources/index.rst.txt)). I tried disabling these styles locally in devtools, to no visible effect, but perhaps their mere presence changes the build process for readthedocs somehow? 

Hypothesis issue: https://github.com/HypothesisWorks/hypothesis/issues/4347

humitos:
Closes #584

### DIFF
--- a/tests/linkpreviews.test.html
+++ b/tests/linkpreviews.test.html
@@ -39,7 +39,7 @@
           // Restore the fake server to its original state
           server.restore();
 
-          document.querySelectorAll("div.tooltip").forEach(el => el.remove());
+          document.querySelectorAll("div.tooltip").forEach((el) => el.remove());
         });
 
         describe("Link previews tests", () => {


src/linkpreviews.js
```diff
@@ -189,7 +189,7 @@ function setupTooltip(el, doctoolname, doctoolversion, selector) {
       );
       newTooltip.setAttribute(TOOLTIP_DATA_HREF, anchorElement.href);
       newTooltip.classList.add("tooltip");
-      anchorElement.insertAdjacentElement("afterend", newTooltip);
+      document.body.insertAdjacentElement("beforeend", newTooltip);
```
agjohnson, line 192: I am wondering what effect this might have. I assumed looking at this code that the element was inserted adjacently for positioning purposes. For example, if the tooltip is positioned relatively, measuring from the document body would be different than the measurement from a sibling element to the tooltip trigger element.

This might be noticeable if the link was inside an element that had a large `margin-left` for example, as the document body would not have this.

But as far as I can tell this doesn't look like it should be the case?

agjohnson (approved):
I'm mostly confident what I pointed out is not a problem but it's hard to say definitively. If we see in testing that any tooltips are not lining up with the link trigger elements, that would confirm my suspicion.

PR #596, opened 2025-06-03
VisualDiff: complex selector and `?maincontent=` API argument

### LINKED ISSUE #576: Visual diff is not working on mkdocs, double-selects root element
## Details

* Read the Docs project URL: https://prairielearn--11830.org.readthedocs.build/en/11830/question/index.html?readthedocs-diff=true
* Config: https://github.com/PrairieLearn/PrairieLearn/blob/master/.readthedocs.yml

## Expected Result

When I click visual diff, it diffs correctly. Instead, it currently throws an error.

Image

## Actual Result

```js
this.performDiff(this.cachedRemoteResponse.content);
// ...

    const htmlDocument = parser.parseFromString(remoteContent, "text/html");
    const oldBody = htmlDocument.documentElement.querySelector(
      this.rootSelector,
    );
```
The JS fetches this url: https://prairielearn--11830.org.readthedocs.build/_/api/v3/embed/?url=https%3A%2F%2Fprairielearn.readthedocs.io%2Fen%2Flatest%2Fquestion%2Findex.html&maincontent=main+%3E+div+%3E+div.md-content

The content it fetches is at the root selector:

Image
Image

It then tries to select the root node of the fetched document, and fails, since it has already fetched it!

Image

## Fix: don't select content at the root selector once it has been fetched through the embed link.

humitos:
When we send the `?maincontent=` CSS selector to the API endpoint, the response file is chunked to only return the chunk that contains the CSS selector.

There are cases when we use inner CSS selector that won't work on the returned response (eg. `main > div > div.md-content`) since the response will contain only the `<div class="md-content">...` part.

In those cases, the query for `this.rootSelector` will fail and we will fallback to the `document.body.firstElementChild` which will always be the element we are looking for.

Unfortunately, I wasn't able to write a test case for this. I've tried different approaches and wasn't able to :/

Closes #576

### DIFF
--- a/src/docdiff.js
+++ b/src/docdiff.js
@@ -186,9 +186,21 @@ export class DocDiffElement extends LitElement {
   performDiff(remoteContent) {
     const parser = new DOMParser();
     const htmlDocument = parser.parseFromString(remoteContent, "text/html");
-    const oldBody = htmlDocument.documentElement.querySelector(
-      this.rootSelector,
-    );
+
+    // We first try to get the `rootSelector` from the `remoteContent`.
+    // However, depending on how the selector is constructed, it may not exist
+    // even if the response is valid.
+    //
+    // This happens when we send `?maincontent=` to the API backend with a
+    // complex selector (eg. "main > div > div.md-content") since in the
+    // response the first elements (e.g. "main > div") won't exist because the
+    // content is already parsed to return only the `?maincontent=` selector.
+    //
+    // In those cases, we always pick the `firstElementChild` of the body.
+    const oldBody =
+      htmlDocument.documentElement.querySelector(this.rootSelector) ||
+      htmlDocument.documentElement.querySelector("body").firstElementChild;
+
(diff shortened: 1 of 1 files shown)


humitos:
This is a simple change that I'd like to merge and deploy if possible.

ericholscher (approved):
This feels a bit odd to only have for `oldBody` -- should we have similar logic for the actual diff content as well? I don't fully understand the context of this change, but it feels like a reasonable default.

humitos:
> This feels a bit odd to only have for `oldBody` -- should we have similar logic for the actual diff content as well?

No. We need this _only_ in the old body because this comes from the API and it's a chunked part of the HTML. The current body is the whole HTML and the selector will work there.

The chunked one (old body) is wrapped between `html` and `body` elements automatically when calling `DOMParser().parseFromString()`. That's why we need this default here when the selector is not found in the old body.

PR #604, opened 2025-06-20
Reintroduce the stickybox on wide screens

davidfischer:
Alternates between using the stickybox placement and the footer placement on unknown themes. It also adds an ID to the stickybox placement so we can see performance of it.

As written, this introduces a random element which may not be desirable. We could instead introduce the stickybox only on very wide screens (say >=1600px) or we could base it on the title of the project or something like that to make it deterministic. I'm open to suggestions.

>         // TODO: Check this placement on the dashboard,
>         // and see how this is performing.

The footer placement is performing very well at ~4x the CTR with ~50% the view rate vs. previously.

Ref: https://github.com/readthedocs/addons/pull/562

### DIFF
--- a/src/ethicalads.js
+++ b/src/ethicalads.js
@@ -198,11 +198,22 @@ export class EthicalAdsAddon extends AddonBase {
         if (elementToAppend) {
           elementToAppend.append(placement);
         }
+      } else if (window.innerWidth > 1300 && Math.random() > 0.5) {
+        // https://ethical-ad-client.readthedocs.io/en/latest/#stickybox
+        placement.setAttribute("data-ea-type", "image");
+        placement.setAttribute("data-ea-style", "stickybox");
+        placement.setAttribute("id", "readthedocs-ea-image-stickybox");
+        this.addEaPlacementToElement(placement);
+        // `document.body` here is not too much relevant, since we are going to
+        // use this selector only for a floating stickybox ad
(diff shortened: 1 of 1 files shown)


humitos (approved):
Looks good to me. I tested it locally and it worked great!

src/ethicalads.js
```diff
@@ -198,11 +198,22 @@ export class EthicalAdsAddon extends AddonBase {
         if (elementToAppend) {
           elementToAppend.append(placement);
         }
+      } else if (window.innerWidth > 1300 && Math.random() > 0.5) {
```
agjohnson, line 201: This could be another place to eventually use media query events, so that we can re-evaluate this logic if the view port size changes. This will just execute once on page load.

Noting for later, I don't think we are using this type of logic anywhere else yet.

agjohnson (approved):
Looks good :+1:

src/ethicalads.js
ericholscher, line 201: I'd like to avoid the randomness here. I think a better approach is using the same logic for each user, so it doesn't change randomly on refresh.

ericholscher (approved):

ericholscher:
Will merge this for release today.

ericholscher:
Did a simple version here: https://github.com/readthedocs/addons/pull/607

PR #613, opened 2025-07-02
Fix issue with null current build validation

agjohnson:
When a project is private and version private the current build can be null. This was presenting as an exception in validation, especially for the notification addon.


agjohnson:
Going to merge this so I can get out a release but @humitos you should still look.

agjohnson:
- [ ] Validation: make sure a null build.current is used correctly

PR #614, opened 2025-07-02
Move release doc to our operations playbook.

agjohnson:
This moves our release operations to live alongside the rest of our release
docs.

- https://github.com/readthedocs/readthedocs-ops/pull/1670


docs/changelog.rst
humitos, line 1: We should keep this file or create a redirect to https://github.com/readthedocs/addons/releases

docs/releasing.rst
humitos, line 1: We should create a redirect for this URL as well.

docs/changelog.rst
agjohnson: Redirect is better, I doubt anyone is using these docs. I don't have admin so you'll have to add.

humitos: I sent you an invitation.

PR #624, opened 2025-07-22
EthicalAd: switch between light/dark mode on Furo

humitos:
Initial implementation to try out ad in dark mode on ethicalads:

- It shows the ad in the defined style at the beginning
- It listens to changes in the theme mode and switch ad theme as well
- It only works on furo for now
- The structure will be the same for other themes, we just need to implement minor details on them

### Sidebar ad

Note that I had to use `style=image` instead of `style=readthedocs-sidebar` because the latter doesn't support dark mode. We should probably implement this in the ad client, and switch it back here.

[Peek 2025-07-22 11-39.webm](https://github.com/user-attachments/assets/7c4d272c-d13e-44b7-a954-c142b148691e)

### Text ad at the bottom of the page

[Peek 2025-07-22 11-40.webm](https://github.com/user-attachments/assets/e01d8718-85c9-44ff-9bf9-641f5e501766)

### Fixed footer text ad

Note this is not adding in this PR because it needs this chunk of code to work: [`d7d3c9e` (#618)](https://github.com/readthedocs/addons/pull/618/commits/d7d3c9e3d0d4f0f8f05f5fa0ef82f3c385c130d6#diff-10e47ee23788d0e8d3d4e13cee726fd04c8fe035922a74ac1c7d432db1dfb9faR102). If we want to have it working, I can bring that code in here.

[Peek 2025-07-22 11-46.webm](https://github.com/user-attachments/assets/9630eab2-3da1-469c-aa1a-a80c6e0d85ce)

Related readthedocs/meta#192


humitos:
@ericholscher should we go ahead and try out this implementation? can you take a look at the code and do a review?

davidfischer:
> @davidfischer is this correct? does `readthedocs-sidebar` not support dark mode?

Correct. This is only implemented on the newer themes.

We have an [open issue on the ad client](https://github.com/readthedocs/ethical-ad-client/issues/224) where we discuss having styles that are adaptive based on `data-theme` but I think it should also handle things like `body.dark` and `html.dark`. I'm leaning toward implementing that sooner than later.

ericholscher (approved):
Good points -- I'm :+1: on moving forward here then, and seeing how it goes. I'm assuming these watchers don't add significant overhead to the page or anything.

The only other worry is the ad format changing slightly -- but I don't think that should have too much impact, but something to monitor.

humitos:
Cool! @ericholscher want to do a review on #627 that implements the generic approach?

humitos:
> I was suggesting doing this in the ad client CSS, not in our JavaScript ([readthedocs/ethical-ad-client#224](https://github.com/readthedocs/ethical-ad-client/issues/224)).

I opened a proposal for this at https://github.com/readthedocs/ethical-ad-client/pull/227 cc @ericholscher @davidfischer

PR #635, opened 2025-08-27
EA: use `adaptive-css` feature for light/dark model

humitos:
Let EA to handle light/dark mode. We keep updating the internal variable `documentationThemeMode` just in case we need it in the future for now. We can remove it eventually if we want.

Requires https://github.com/readthedocs/ethical-ad-client/pull/229

### DIFF
--- a/src/ethicalads.js
+++ b/src/ethicalads.js
@@ -345,6 +345,8 @@ export class EthicalAdsAddon extends AddonBase {
     if (
       window.location.hostname === "docs.readthedocs.com" ||
       window.location.hostname.endsWith(".devthedocs.org") ||
+      // Use new beta client on Furo like themes for now.
+      // This allows us to test the dark/light mode.
       docTool.isSphinxFuroLikeTheme()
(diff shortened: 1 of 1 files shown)


davidfischer:
The adaptive light/dark mode is available on the beta client and I'll aim to roll it out more widely in a couple days.

Based on this code, it looks like your plan is to always set the placement as adaptive and then it will adapt if the theme supports it. That should be OK in virtually all cases.

humitos:
> Based on this code, it looks like your plan is to always set the placement as adaptive and then it will adapt if the theme supports it. That should be OK in virtually all cases.

Yes. I will start there relying completely con the EA CSS. We can make adjustments (enable/disable this) for specific themes/doctools in the near future if we find that it doesn't work as expected. It worked in all the themes I tried, so far.

ericholscher (approved):
Excited for this to move forward! I think this will make the ads way nicer for a lot of RTD users. 💯

PR #641, opened 2025-09-30
EthicalAd: add support for fixed footer ad in all the themes

humitos:
This is the logic we follow to show the fixed footer ad:

* We detect the doctool/theme using our common heuristic.
* The image nav bar ad is below the fold.
* Inject the fixed footer ad on a specific selector (if needed, eg. "body").
* Add `padding-bottom` to specific CSS selectors to not cover content.

### VitePress

Screenshot_2025-09-30_13-14-45

### Material for MkDocs

Screenshot_2025-09-30_13-15-06

Closes https://github.com/readthedocs/meta/issues/192

### DIFF
--- a/src/ethicalads.js
+++ b/src/ethicalads.js
@@ -154,10 +154,12 @@ export class EthicalAdsAddon extends AddonBase {
 
           placement.setAttribute("data-ea-type", "readthedocs-sidebar");
           knownPlacementFound = true;
-        } else {
-          fixedFooterAdSelectors = ["footer.md-footer"];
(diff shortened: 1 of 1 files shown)


humitos:
> I only worry about Mkdocs Material since they have been upset with us, so if this ends up covering their branding

I tested it on purpose because of this and it works fine. However, I'm 👍🏼 to remove it from now if we don't feel comfortable with this change.

humitos:
I disabled it for now. We can uncomment the code when we feel ready for it.

Issue #574, opened 2025-04-16
New addons UI


humitos:
We discussed a lot about addons at the offsite and we came back with a great and more clear idea on how the UI should be. I'm posting here a small video that briefly shows how it:

https://github.com/user-attachments/assets/ebc4f8a8-0454-4db5-b5ac-a424ea57c34d

There are a few **key points** we want to highlight here:

- The UI element will be at the top center
- The UI element will auto-hide after a few seconds
- There is no "minimized" and "expanded" versions of each addon (we realized that was not needed)
- Version and language selectors are dropdowns
- Search will show the results in a "expanded version" of the same UI (similar to Algolia, Material for MkDocs and others)
- Versions can be pinned from the dashboard and they will be shown at the top of the version list
- Notifications will be shown stacked at the bottom of the UI
- Notifications will auto-hide after a few seconds
- There will be a clickable notification icon to show all the notifications
- The default version is highlighted with a "branch + star" icon

Here is the thought process we went through. Click to expand details and see all the drawings:

<details>

### humitos

|Drawings||
|---|---|
|![Image](https://github.com/user-attachments/assets/4fd32bd2-1451-4439-a950-b3e2c5210ebf)|![Image](https://github.com/user-attachments/assets/b0b4c8e1-d40f-46b1-8078-8252f3bbddfa)|
|![Image](https://github.com/user-attachments/assets/e1f734e8-e26b-4db2-a00f-4af37cb2b85b)|![Image](https://github.com/user-attachments/assets/c517eb9e-b7ec-4be4-b2d2-a99c3b9d664a)|

### agjhonson

|Drawings||
|---|---|
|![Image](https://github.com/user-attachments/assets/d843a4ae-a244-4cf5-9b95-c2f598e706e6)|![Image](https://github.com/user-attachments/assets/99cc1fa8-1641-461a-868a-0eb481966fe8)|
|![Image](https://github.com/user-attachments/assets/58101f41-a6f5-47c9-bb91-396652af5067)|![Image](https://github.com/user-attachments/assets/bb922204-48a4-4d03-977a-c556a71928f1)|
|![Image](https://github.com/user-attachments/assets/f8dfbae5-9f44-4e0b-bc36-9bdd478ac54e)|![Image](https://github.com/user-attachments/assets/92cf7ebe-5af7-464b-8aee-d63b16c8962f)|
[...]

PR #688, opened 2026-04-10
New addons UI: top-center bar with panel-based navigation

ericholscher:
## Summary

New addons UI redesign base. See #574 for design mockups and video.

**Toggle:** `?readthedocs-flyout-v2=true`

## Design

The flyout is a single horizontal bar that shows the active addon's UI inline. No expanded pane — the bar switches content based on which addon is selected via the hamburger.

### Menu (default) — version & language dropdowns
Menu bar

### Hamburger — switch between addons
Hamburger

### Search — inline input
Search

### File diff — inline controls
Filediff

### Settings — config placeholder
Config

## Files

| File | Purpose |
|---|---|
| `src/flyout-v2.js/css` | Bar with hamburger + inline addon views |
| `src/search-panel.js/css` | Search component |
| `src/filetreediff-panel.js/css` | File diff component |
| `src/defaults.js` | Registers v2 CSS variables |
| `src/flyout.js` | Original (unchanged except v2-param guard) |

https://claude.ai/code/session_01GS43pv3hBjSE8BBLnyCSTE


humitos (commented):
I pull this PR down and I tested it locally to feel the UX by myself. I also took a look at the code and I can say the JS code is probably pretty close to what I could write by myself, so in that regard is 👍🏼  There isn't anything new here and it seems it follows the patterns we have established.

On the other hand, the CSS is code is where I'm not able to evaluate this PR --and there is a ton of it because it's the most important part of what we want to build for the new addons UI.

That said, we are not covering what we designed here. Mainly the hamburger menu with the buttons and panes. Those buttons should make the whole UI to switch to that particular feature as shown in the video we have in the issue.

I'm not sure how much we can guide the AI to move in that direction, but it would be awesome if we can make the AI to build the HTML/CSS we've designed together in the offsite since the JS is pretty close to what we need I'd say --I mean, I would feel comfortable updating the JS code by myself with any required tweak, but not the HTML/CSS.

PR #701, opened 2026-06-01
Flyout: fix crash in `getLinkWithFilename` and `addUtmParameters` when React hydration removes meta tags

ericholscher:
## Summary

Fixes `TypeError: Cannot read properties of undefined (reading 'replace')` in `getLinkWithFilename`.

**Root cause:** When `resolver.filename` is `null` in the API response (e.g. pyrit.readthedocs.io), the code falls into the metadata fallback and calls `getMetadataValue("readthedocs-resolver-filename")`. On React-hydrated pages (see #278), hydration wipes the Cloudflare-injected meta tag, so `getMetadataValue` returns `undefined`. The code then falls through and calls `.replace()` on `undefined`, crashing.

**Fix:** Guard against `resolverFilename` still being falsy after the metadata fallback, falling back to the base URL as-is.

## Test plan

- [x] New unit test: `getLinkWithFilename(url, null)` with no meta tag in DOM returns the base URL without crashing
- [x] All 6 `getLinkWithFilename` unit tests pass
- [ ] Verify flyout renders without errors on a React-hydrated page with `resolver.filename: null`

Related: #278

---
*Generated by Copilot*

https://claude.ai/code/session_01Rq8JDpWyd6HUjz75X1qRyY


humitos (commented):
This new approach looks better.

I want to pull down this PR and test it over https://pyrit.readthedocs.io/0.13.0/ to confirm it works as expected.

src/utils.js
```diff
@@ -376,6 +376,13 @@ export function getLinkWithFilename(url, resolverFilename) {
     }
   }
 
+  // resolverFilename may still be undefined if the Cloudflare-injected meta
+  // tag was removed by React hydration (see issue #278); fall back to the
+  // base URL in that case.
+  if (!resolverFilename) {
```
agjohnson, line 382: Is this supposed to be testing for undefined or truthy? Comment above implies undefined.


```suggestion
  if (resolverFilename == undefined) {
```

This logic will currently trigger if path is a false looking string, `""`:

```javascript
> !Boolean("")
true 
```

humitos:
I updated the code to handle hydrated sites in a better way. I also added auto-detection for mystmd doctool.

I tested this locally using PyRIT as example and it worked fine 👍🏼

humitos (approved):

ericholscher:
@humitos I don't see the update in the PR now that you mention it. Did you push it up?

humitos:
meh, I'm dumb. I didn't push my code. I opened https://github.com/readthedocs/addons/pull/705 as a follow up of this with my changes.

PR #704, opened 2026-06-03
Ads: handle Sphinx piccolo theme

### LINKED ISSUE #13055: Ad injection code messes up banner style when it mistakes Piccolo theme for RTD's own theme
## Details
The [Piccolo theme](https://piccolo-theme.readthedocs.io/) for sphinx loads its JS from `_static/js/theme.js`. Unfortunately, this lines up exactly with readthedocs.org's own `sphinx_rtd_theme`.

Because of this, `isSphinxReadTheDocsLikeTheme()` in the ad-injection code returns `true`. It then looks for the RTD sidebar, which it of-course doesn't find and falls back to dumping the footer ad's height onto various elements as `padding-bottom`. Text gets covered up by the now-oversized elements.

See Piccolo theme employed by these projects and watch the top bar extend over the docs text:
* https://app.readthedocs.org/projects/vapoursynth-analog/
  * Live site: https://vapoursynth-analog.justinarthur.com/en/latest/
* https://app.readthedocs.org/projects/vsfieldkit/
  * Live site: https://vsfieldkit.justinarthur.com/en/latest/ 

Coincidentally, the issue occurs on the Piccolo theme's own documentation linked above. This has been happening for at least a couple months now.

## Expected Result
The ad-injection code leaves banner elements and other critical parts of the theme alone.

## Actual Result
Many elements extend below where they were supposed to finish and overlap text making it unreadable.

humitos:
### Sidebar

Screenshot_2026-06-03_09-20-55

### Footer

Screenshot_2026-06-03_09-21-05

Closes https://github.com/readthedocs/readthedocs.org/issues/13055

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -4,6 +4,7 @@ import {
   SPHINX,
   SPHINX_FURO,
   SPHINX_ALABASTER,
+  SPHINX_PICCOLO,
   SPHINX_READTHEDOCS,
   SPHINX_IMMATERIAL,
   SPHINX_PYDATA,
@@ -601,7 +602,9 @@ export class DocumentationTool {
       this.documentationTool || this.getDocumentationTool();
 
     if (documentationTool === SPHINX) {
-      if (this.isSphinxAlabasterLikeTheme()) {
+      if (this.isSphinxPiccoloTheme()) {
+        return SPHINX_PICCOLO;
+      } else if (this.isSphinxAlabasterLikeTheme()) {
         return SPHINX_ALABASTER;
       } else if (this.isSphinxReadTheDocsLikeTheme()) {
         return SPHINX_READTHEDOCS;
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -772,6 +776,16 @@ export class DocumentationTool {
     return false;
   }
 
+  isSphinxPiccoloTheme() {
+    if (
+      document.querySelectorAll('script[src*="_static/js/petite-vue.js"]')
```
ericholscher, line 781: This seems like a vue library include, not picalo specific?

humitos: Yeah, but it's the only thing I found that is different from the other themes. That file is only used in this theme. 

We don't have strong heuristic here because there are no too much data in the HTML

ericholscher (approved):
This seems fine for now, but will likely break something else in the future :/

humitos:
> This seems fine for now, but will likely break something else in the future :/

Yeah, but that's by design 🙃 -- we already knew that this type of heuristic is a not perfect 😞 

I built it on test-builds at https://test-builds.readthedocs.io/en/piccolo/ and added into the addons examples to keep an eye on this theme.

PR #705, opened 2026-06-04
flyout: fix crash when resolver filename is undefined

humitos:
This is a follow up of https://github.com/readthedocs/addons/pull/701 because I'm dumb and I didn't push my code.


ericholscher (approved):
This approach makes sense to me.

PR #716, opened 2026-09-01
ethicalads: follow Material/Zensical dark mode

ericholscher:
Ads on Material for MkDocs and Zensical sites always render in light mode, even when the page itself is dark — reported on https://goss.readthedocs.io/en/stable/ (Zensical with an auto light/dark palette) with a dark system preference.

Root cause: we add the `adaptive-css` class to the placement, but the EthicalAds client's adaptive mode only reacts to markers it knows on `html`/`body` — the `dark`/`auto` classes, `data-theme`, or `data-bs-theme`. Material-family themes (Material for MkDocs, Zensical, sphinx-immaterial) signal dark mode with `data-md-color-scheme="slate"` instead, so no selector ever matches and the ad keeps the light default.

**Updated approach:** the ad client now supports this directly. [readthedocs/ethical-ad-client#243](https://github.com/readthedocs/ethical-ad-client/pull/243) (released in v2.5.0) added a [`data-ea-dark-selector`](https://ethical-ad-client.readthedocs.io/en/latest/#custom-dark-selector) attribute: the client evaluates the selector and watches `html`/`body` attribute changes itself, toggling its `dark` class on the placement. So instead of the hand-rolled `MutationObserver` from the first revision, this PR now:

- Sets `data-ea-dark-selector="body[data-md-color-scheme='slate']"` on the placement when the page uses Material's color scheme attribute. Themes without the attribute are unaffected.
- Loads the **beta** client on those pages, since v2.5.0 is only on the beta CDN so far (the stable CDN still serves v2.4.0). This follows the existing pattern for Furo-like themes and can be dropped once v2.5.0 is promoted to stable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FKD7pkDGWQCQCpXNaRqwPR

### DIFF
--- a/src/ethicalads.js
+++ b/src/ethicalads.js
@@ -312,6 +312,7 @@ export class EthicalAdsAddon extends AddonBase {
     if (placement !== null) {
       // Allow EA to switch between light/dark mode
       placement.classList.add("adaptive-css");
+      this.syncDarkModeWithPageColorScheme(placement);
 
       // This ensure us that all the `data-ea-*` attributes are already set in the HTML tag.
       placement.setAttribute("data-ea-manual", "true");
@@ -389,6 +390,35 @@ export class EthicalAdsAddon extends AddonBase {
     }
 
     return placement;
+  }
+
+  syncDarkModeWithPageColorScheme(placement) {
+    // Material for MkDocs, Zensical, and sphinx-immaterial signal their color
+    // scheme with a `data-md-color-scheme` attribute ("slate" is dark), which
+    // the EthicalAds client doesn't recognize in its `adaptive-css` mode.
+    // Toggle the client's explicit `dark` class to match the page.
+    const element = document.querySelector(
+      "html[data-md-color-scheme], body[data-md-color-scheme]",
+    );
+    if (!element) {
+      return;
+    }
+
+    const updateDarkClass = () => {
+      if (element.getAttribute("data-md-color-scheme") === "slate") {
+        placement.classList.add("dark");
+      } else {
+        placement.classList.remove("dark");
+      }
+    };
+    updateDarkClass();
+
+    // Keep the ad in sync when the user or the OS switches modes.
+    const observer = new MutationObserver(updateDarkClass);
+    observer.observe(element, {
+      attributes: true,
+      attributeFilter: ["data-md-color-scheme"],
+    });
   }
 
   elementAboveTheFold(element) {
--- a/tests/ethicalads.test.js
+++ b/tests/ethicalads.test.js
@@ -1,5 +1,64 @@
-import { expect, assert, fixture, html } from "@open-wc/testing";
+import { expect, assert, fixture, html, aTimeout } from "@open-wc/testing";
 import { EthicalAdsAddon } from "../src/ethicalads";
+
+describe("EthicalAds addon dark mode", () => {
+  const config = {
+    addons: {
+      ethicalads: {
+        enabled: true,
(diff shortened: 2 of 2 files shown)


humitos:
> Root cause: we add the `adaptive-css` class to the placement, but the EthicalAds client's adaptive mode only reacts to markers it knows on `html`/`body` — the `dark`/`auto` classes, `data-theme`, or `data-bs-theme`. Material-family themes (Material for MkDocs, Zensical, sphinx-immaterial) signal dark mode with `data-md-color-scheme="slate"` instead, so no selector ever matches and the ad keeps the light default.

It seem this should be implemented in ethical ad client instead so the `adaptive-css` works out of the box, right?

davidfischer:
One possibility since there is an unlimited number of attributes where the light/dark mode could be set is to set the attribute on the ad placement (eg. `data-ea-adaptive-attribute`). Then the ad client could monitor that attribute for changes and trigger some sort of update code.

ericholscher:
Could be in either, we should just fix it somewhere because it currently looks pretty bad :( 

I really wish there were standards here. :(

davidfischer:
Unfortunately, there isn't a standard here.

Looking more closely at what Zensical is doing, they're setting `data-md-color-scheme="slate"` or `data-md-color-scheme="default"`. Our normal `light`/`dark` selectors wouldn't even work. There's a few possibilities:

1. We could just handle Zensical's settings. We would just add `body[data-md-color-scheme="slate"]` [here](https://github.com/readthedocs/ethical-ad-client/blob/19661dce22dd0f74bba9c4a1ee6c7fe18ae24477/styles.scss#L240-L245). This is straight-forward but a one-off fix for them.
2. We add an attribute like `data-ea-dark-selector="body[data-md-color-scheme='slate']"`. The client sets up a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to watch for changes to that selector. When it changes, we'd have logic to evaluate whether the ad should be dark mode or not and then apply the class manually. It's more work but it would allow this to work on any theme as long as that attribute is set. I added a ~not fully tested draft~ implementation [here](https://github.com/readthedocs/ethical-ad-client/pull/243).

davidfischer:
This change is live on the beta version of the ad client if you want to give it a try. Use the [beta client](https://media.ethicalads.io/media/client/beta/ethicalads.min.js) and set `data-ea-dark-selector="body[data-md-color-scheme='slate']"` and it should toggle correctly.

ericholscher:
This is updated to use the new approach ✨ 

Screenshot 2026-09-14 at 9 54 09 AM

src/ethicalads.js
```diff
@@ -448,7 +465,9 @@ export class EthicalAdsAddon extends AddonBase {
       window.location.hostname.endsWith(".devthedocs.org") ||
       // Use new beta client on Furo like themes for now.
       // This allows us to test the dark/light mode.
-      docTool.isSphinxFuroLikeTheme()
+      docTool.isSphinxFuroLikeTheme() ||
+      // `data-ea-dark-selector` requires client v2.5.0, only on beta so far.
+      this.usesMaterialColorScheme()
     ) {
       src = "https://media.ethicalads.io/media/client/beta/ethicalads.min.js";
```
davidfischer, line 466: This should no longer be necessary. The regular client has the new `data-ea-dark-selector` option now.

davidfischer (approved):

